PR #60, opened 2015-04-29
update doc of part on using file-stream-rotator

ytfei:
`var createRotatingStream = require('file-stream-rotator').getStream` will return a detached function without expected object context, then if we invoke `createRotatingStream`, error will be thrown cause `this` is undefined.


dougwilson:
Thank you!

PR #62, opened 2015-05-07
Adding method to get formatted string inside format function. Closes #61

### LINKED ISSUE #61: Is there a way to get the formatted output string inside the format function?
Im looking for a way to get the formatted string that morgan would otherwise send to the console, so I can pass it and an object representing the data to my logger. Currently I am just building the object with tokens['token'] and using that object to build up a string to output, but it is rather ugly and hard to match the format that morgan outputs on its own.

JimtotheB:
This is a fairly simple method that uses the existing compile function to return a formatted string. I also added 3 tests, and added a section to the docs detailing the "function as format" option.

This closes issue #61

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,15 +200,11 @@ exports.format = function(name, fmt){
  * @param {string} format
  * @returns {string}
  */
+
 exports.getFormatted = function(req, res, format) {
-  if(!req['headers']){
-    deprecate('getFormatted expects a req and request object as parameters.')
-    return ''
-  }
   var fmt = (exports[format] || format || exports.default)
   var formatted = compile(fmt);
   return formatted(exports, req, res)
-
 }
 
 /**
--- a/index.js
+++ b/index.js
@@ -193,6 +193,25 @@ exports.format = function(name, fmt){
 };
 
 /**
+ * Returns formatted string. Useful when using a function as a format.
+ *
+ * @param {object} req
+ * @param {object} res
+ * @param {string} format
+ * @returns {string}
+ */
+exports.getFormatted = function(req, res, format) {
+  if(!req['headers']){
+    deprecate('getFormatted expects a req and request object as parameters.')
+    return ''
+  }
+  var fmt = (exports[format] || format || exports.default)
+  var formatted = compile(fmt);
+  return formatted(exports, req, res)
+
+}
+
+/**
  * Apache combined log format.
  */
 
--- a/README.md
+++ b/README.md
@@ -287,6 +287,42 @@ function assignId(req, res, next) {
   next()
 }
 ```
+### using a function as a format.
+Sending log data to a separate logger. 
+
+```javascript 
+var express = require('express')
+var morgan = require('morgan')
+var logger = require('myCustomLogger')
+var app = express()
+
+
+app.use(morgan(function(tokens, req, res){
+
+	// Extract your data from the request.
+	var reqData = {
+     "remote-addr": tokens['remote-addr'](req, res),
+     "date": tokens['date'](req, res),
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -193,6 +193,25 @@ exports.format = function(name, fmt){
...
+ * Returns formatted string. Useful when using a function as a format.
+ *
+ * @param {object} req
+ * @param {object} res
+ * @param {string} format
+ * @returns {string}
+ */
+exports.getFormatted = function(req, res, format) {
+  if(!req['headers']){
+    deprecate('getFormatted expects a req and request object as parameters.')
```
dougwilson: A new function would not have deprecated behavior. If this is to catch a mistake, it should throw instead.

dougwilson:
The concept itself in the PR is sound.

PaperElectron:
Well, It is a runtime error, and a bunch of things can go wrong with a request. Is there a better way to fail gracefully from what amounts to a utility function?

dougwilson:
Call the function in a `try` block.

PaperElectron:
Try - catch would be the obvious solution, Heh, It's really late here. Am I ok to check for req.headers and res.headers to verify that I actually have the correct objects? Also, I'm returning the error as the output of the function if it catches an error. Do you see any problems with this?

dougwilson:
I'd have to see the implementation to really comment :) But the `try`-`catch` would not be in this module, but rather in your calling code if you are concerned you don't know what type of objects you are passing in--this module, if it does any argument validation, should throw `TypeError`.

PaperElectron:
Ok, I'm way over thinking it now. All of the formatting functions already deal with the req, and res objects, so nothing breaks if you pass them junk and nothing this is calling has side effects. 

I really hate to throw on a poorly formatted object passed at runtime. What about just returning the string as morgan outputs it, whatever that may be?

dougwilson:
If you're suggesting to just completely remove the `if(!req['headers']){` that's A-OK with me. if you're suggesting something else, just update the PR and I'll see what you mean :)

PaperElectron:
Would it be possible to get this merged, or closed if not?

dougwilson:
You're welcome to close it, but I do want to merge it. Until July, when Express 3.x drops, these modules' minors are tied to Express 3.x minors. A new feature is a minor, which is the only reason this PR hasn't been merged yet.

PaperElectron:
Ah, I hadn't thought about the downstream implications. Im fine with just letting this hang open until you are ready to move on it, I just have a selfish interest in the functionality I added. I will just use my fork in my project in the meantime.

dougwilson:
There is still work being done here, but in the next version, you can at least accomplish this task using the `morgan.compile` method, in combination with access the needed format string.

dougwilson:
For example, with 1.6.0, your example in the PR could become the following:
```javascript 
var logger = require('myCustomLogger')
var app = express()

var getCommonLine = morgan.compile(morgan.common)

app.use(morgan(function(tokens, req, res){

	// Extract your data from the request.
	var reqData = {
     "remote-addr": tokens['remote-addr'](req, res),
     "date": tokens['date'](req, res),
     "method": tokens['method'](req, res),
     "status": tokens['status'](req, res),
     "response-time": tokens['response-time'](req, res)
	}
	
	// a formatted string.
	var formatLine = getCommonLine(morgan, req, res)
	
	logger.log(formatLine, reqData)
	
	//or do something with your data and return the formatted string.
	logger.writeData(reqData)
	return formatLine
	
	//or do something async and write your formatted line manually.
	logger.writeAsyncData(reqData, function(err){
		console.log(formatLine)
	})
});
```

PR #63, opened 2015-05-21
Write stream includeTokens

blahed:
We had a need to access processed tokens inside the stream write callback. We pass the tokens to winston as metadata and didn't want to double process the tokens.

This change modifies the compile function to return the formatted string and the tokens.

FWIW I think we've been working towards the same thing as #62. We didn't see that PR when we started work on this and we've taken a different approach that avoids calling the token functions twice.

One potential downside (or maybe upside) to this approach is that your tokens need to be in the format string for them to be available in the write callback.

```javascript
morgan('common', {
  stream: {
    includeTokens: true,
    write: function(line, metadata) {
      logger.info(line, metadata);
    }
  }
})
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,8 +70,9 @@ exports = module.exports = function morgan(format, options) {
   // steam
   var buffer = options.buffer
   var stream = options.stream || process.stdout
-  var includeTokens = options.stream ? options.stream.includeTokens === true : false
-
+  // logger
+  var logger = options.logger
+  
   // buffering support
   if (buffer) {
     deprecate('buffer option')
@@ -105,7 +106,7 @@ exports = module.exports = function morgan(format, options) {
     };
   }
 
-  return function logger(req, res, next) {
+  return function(req, res, next) {
     req._startAt = process.hrtime();
     req._startTime = new Date;
     req._remoteAddress = getip(req);
@@ -122,7 +123,7 @@ exports = module.exports = function morgan(format, options) {
       
       if (typeof formatted === 'object') {
         line = formatted.line;
-        tokenPairs = formatted.tokenPairs
+        tokenPairs = formatted.tokenPairs;
       } else {
         line = formatted;
       }
@@ -133,7 +134,11 @@ exports = module.exports = function morgan(format, options) {
       }
 
       debug('log request');
-      includeTokens ? stream.write(line + '\n', tokenPairs) : stream.write(line + '\n');
+      if (logger) {
+        logger(line, tokenPairs);
+      } else {
+        stream.write(line + '\n');
+      }
     };
 
     // immediate
--- a/index.js
+++ b/index.js
@@ -70,6 +70,7 @@ exports = module.exports = function morgan(format, options) {
   // steam
   var buffer = options.buffer
   var stream = options.stream || process.stdout
+  var includeTokens = options.stream ? options.stream.includeTokens === true : false
 
   // buffering support
   if (buffer) {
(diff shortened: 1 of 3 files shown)


dougwilson:
The biggest problem with your example, is that `stream` can only be a Node.js stream instance; you should never be passing in a `write` function and other options using it at all.

index.js
```diff
@@ -115,15 +116,24 @@ exports = module.exports = function morgan(format, options) {
...
+      
       if (null == line) {
         debug('skip line')
         return
       }
 
-      debug('log request')
-      stream.write(line + '\n')
+      debug('log request');
+      includeTokens ? stream.write(line + '\n', tokenPairs) : stream.write(line + '\n');
```
dougwilson: This is not valid, unfortunately. The second argument to `stream.write` already is defined: it's encoding, you cannot use it for something else.

blahed:
Ah, yeah, that's how it was being used in the logger we have here, stream as an object with a function. It actually had a local copy of morgan with a modification to pass the req and res as arguments to the write, which resulted in calculating the tokens twice and looking really awkward – two of the things we wanted to fix.

So what's the recommended way to tie into an existing logger and have access to the tokens? #62 seems less than ideal as well.

dougwilson:
It's hard for me to comment on more appropriate solutions without knowing your actual problem you are trying to solve. For example, is there a way you can explain in greater detail what "We had a need to access processed tokens inside the stream write callback." is?

blahed:
We need to pass the morgan formatted log line to our logger, which happens to be winston. Winston supports a metadata object and we want to pass the tokens and their values as the metadata object.

What's the recommended integration with winston? It seems like the stream option is the only way to get access to the formatted log line, are we missing something?

dougwilson:
Gotcha. You are not missing something, I just wanted to understand the goal is all. Basically right now `morgan` is only written to function as a purge `access.log`-esque. thing, which just writes log lines to a Node.js stream.

Over the past few weeks or so, it seems like there is a demand to get this same functionality, but also get access to the token functions as well, for use in more rich logging scenarios. Does that sound right?

The current PR's major flaw is that with an option enable, we violate the `.write` Node.js stream API, which is `.write(val, encoding, cb)`, where only `val` is required. We need to put more thought into how to support these rich logging scenarios.

blahed:
Yup, you're spot on.

We have a logger, not a stream, and we want the formatted line and the token key/value pairs.

The way morgan plugs into an existing logger isn't super important to us, but the one thing we were concerned with was that the token functions are being called when formatting the string, so we wanted to store the token values at that stage and make them available, vs calling those same functions again somewhere else.

The reason we went with abusing the stream option was because these two things look pretty similar:

```javascript
morgan('common', {
  logger: {
    log: function(line) {
      winston.info(line);
    }
  }
})
```

```javascript
morgan('common', {
  stream: {
    write: function(line) {
      winston.info(line);
    }
  }
})
```

However, maybe this gives you enough:

```javascript
morgan('common', {
  logger: function(line) {
    winston.info(line);
  }
})
```

dougwilson:
Gotcha. And yea, we technically don't even guarantee you'll only get one line per cal to `.write` (I mean, just look at the buffer option), nor do we guarantee that you'll even have complete lines for each call to `.write` (just you happen to in practice, which is what make the abuse of the `.write` function so prevalent).

blahed:
Just a first pass at the idea to see if it clicks.

blahed:
@dougwilson any thoughts on this updated approach?

dougwilson:
I'm still a little iffy on this approach, but not really dismissing it or the idea (thus why it's still open right now).

blahed:
I think http://stackoverflow.com/questions/27906551/node-js-logging-use-morgan-and-winston shows how people are handling the integration, which isn't really ideal.

What part is feeling iffy? I'd be happy to take another approach if you have some ideas.

The `logger` as an option seems like a way to handle it that lines up with the way people are currently solving the problem. I think the important thing is that the parsed tokens are made available to the logger in some way, as they're useful as metadata.

dougwilson:
It's mainly the fact that the function we invoke (`fmt`), which can very well be a user-defined function, will act differently depending on if it returned a string vs an object, i.e. it acts differently based on returned value, magically. It's also confusing when people have a custom `fmt` function, but provide the `logger` option, that `tokenPairs` is going to be `undefined` since everything out there returns a string right now.

The approach you are taking here seems to cater way more towards being backwards-compatible vs. not being confusing to users.

blahed:
Gotcha. Thinking on it.

PR #73, opened 2015-07-17
Update to log response time once res triggered event 'finish'

twhtanghk:
Both nodejs 0.12 and sailsjs moc-res support res.on 'finish'. To log response time for sailsjs web socket request, it is suggested to log response time by listening to response event 'finish' instead of checking response property finished or complete. Please review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 coverage/
 node_modules/
 npm-debug.log
+/.project
--- a/index.js
+++ b/index.js
@@ -139,6 +139,7 @@ function morgan(format, options) {
 
       // log when response finished
       res.on('finish', logRequest)
+      res.on('close', logRequest)
     }
 
     next();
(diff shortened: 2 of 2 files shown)


dougwilson:
Please investigate the failing tests because of this change.

dougwilson:
Hi @twhtanghk , any progress on investigating why your change causes test failures? Also, I didn't note this earlier, but in order to even accept this PR at all, we need to have a test added that will fail without the change and pass with the change, otherwise the change can easily be undone without us noticing a test failure.

twhtanghk:
Fixed the problem by also listening to close event. Please reveiw

dougwilson:
Thanks! It still needs to have a test added that fails without your change and passes with your change. Also, you need to ensure that if both a `close` and a `finish` event is emitted, the request is only logged once and not twice.

In the end, under closer inspection, all you're doing is slowing re-implementing the `on-finished` module we are using, and all to compensate for an incomplete mock library. Please either use real Node.js HTTP request/response objects or work to get the given mock libraries a more accurate mock of what they are trying to mock.

PR #75, opened 2015-07-31
Add default color in 'dev' format

Nitive:
Add `devDefaultColor` option.
```js
app.use(morgan('dev', { devDefaultColor: 90 }));
```
screen shot 2015-07-31 at 20 48 31

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,6 +56,9 @@ var defaultBufferDuration = 1000;
  * @return {Function} middleware
  */
 
+// default color in 'dev' format
+var devDefaultColor;
+
 function morgan(format, options) {
   var fmt = format
   var opts = options || {}
@@ -71,6 +74,8 @@ function morgan(format, options) {
   if (fmt === undefined) {
     deprecate('undefined format: specify a format')
   }
+
+  devDefaultColor = opts.devDefaultColor || 0
 
   // output on request instead of response
   var immediate = opts.immediate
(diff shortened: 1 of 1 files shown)


dougwilson:
Setting a global will be confusing, as now people cannot create two dev Morgan loggers in the same app with different default colors (perhaps they wanted it to be different per vhost, say).

Please make the change non-global.

Also if you can, please add a description to the readme and tests for the feature :)

Nitive:
I have no idea how can I do it without global variable. I don't have access to options inside `morgan.format`

Nitive:
I will enjoy to do it but I need a little tip. I feel I can not see some obvious thing.

dougwilson:
Hi @Nitive ! I don't think there is an existing way, so you would just have to add that as part of the PR :)

ficristo:
@dougwilson @Nitive Can this PR be prioritized? I really miss the gray text...

dougwilson:
Hi @ficristo , the PR is just waiting for it to be implemented as a non-global. You're welcome to ppick it up in a new PR if you need it :)

ficristo:
I looked a bit at the code and ATM I don't see a way to create two different `dev` logger.
So I think, to support the non-global thing, morgan need a total rewrite.
Can we get away simply adding a new `dev-color` format a rearrange a bit this PR?
There will be a global `devDefaultColor` but used only for the new format.
@dougwilson What do you think?

dougwilson:
> ATM I don't see a way to create two different dev logger. So I think, to support the non-global thing, morgan need a total rewrite.

This is not accurate, it is 100% doable in the current API. Please attempt to fit it in without a re-write.

dougwilson:
>  Can we get away simply adding a new dev-color format a rearrange a bit this PR?

Maybe. I'm not 100% sure what that means, but please feel free to make a PR to demonstrate this idea :)

ficristo:
Actually adding a new format will not solve the global thing. I'll try to find some time to fix and complete https://github.com/expressjs/morgan/pull/89

PR #86, opened 2015-10-10
bump dependencies, npm passed

PeterDaveHello:


dougwilson:
Thanks! We don't typically accept dependency bumps without specific reasons for them being needed (rather than they are just outdated--like it fixed some bug, for example). I have pushed up the past commits that have them bumped from a script.

PR #87, opened 2015-10-10
Update .travis.yml

PeterDaveHello:
also test on node.js v4.0/4.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ node_js:
   - "2.0"
   - "2.5"
   - "3.0"
+  - "4.0"
+  - "4.1"
 sudo: false
 before_install:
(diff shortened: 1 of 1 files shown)


dougwilson:
Hi! We don't typically take updates to the Travis CI configuration file unless it fixes a bug or there's more of a reason to test vs. just "well, it's the latest". I haave gone ahead and accepted this PR anyway :)

PeterDaveHello:
But why not test on v4.x? I think 4.2 should be added as it's a LTS version!

dougwilson:
Your PR only had 4.1.

PeterDaveHello:
What about reopen and let me change to v4.2 :smile: v4.2 was not out when I sending this PR.

PR #89, opened 2015-11-17
Convert morgan to be an object instead of a global function

ficristo:
This is a prerequisite for https://github.com/expressjs/morgan/pull/75.
In this PR I removed the deprecated code too, so I think this need a major update.
**--edit**
I changed the API too.
Before: `morgan('dev')`
After: `morgan().getLogger('dev')`

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,21 +14,22 @@
  * @public
  */
 
-module.exports = morgan
-module.exports.compile = compile
-module.exports.format = format
-module.exports.token = token
+module.exports = Morgan
 
 /**
  * Module dependencies.
  * @private
  */
 
+var assert = require('assert')
 var auth = require('basic-auth')
 var debug = require('debug')('morgan')
 var deprecate = require('depd')('morgan')
 var onFinished = require('on-finished')
 var onHeaders = require('on-headers')
+
+var gFormatters = {}
+var gTokens = {}
 
 /**
  * Array of CLF month names.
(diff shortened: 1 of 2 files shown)


dougwilson:
Please change the API back to the Node.js standard of a factory export.

dougwilson:
The README needs to be updated for all the changes you made as well.

index.js
```diff
@@ -97,8 +84,51 @@ function morgan(format, options) {
...
+ * @param {string} name
+ * @param {function} fn
+ * @public
+ */
+Morgan.token = function (name, fn) {
+  gTokens[name] = fn
+}
+
+Morgan.prototype.getLogger = function (format) {
+  assert (
```
dougwilson: This needs to throw `TypeError`s, not `AssertionError`s

index.js
```diff
@@ -97,8 +84,51 @@ function morgan(format, options) {
...
+    format,
+    'Morgan.getLogger(format): format is mandatory'
+  );
+  assert (
```
dougwilson: This needs to throw `TypeError`s, not `AssertionError`s

ficristo:
> Please change the API back to the Node.js standard of a factory export.

Could you clarify, maybe with an example?

dougwilson:
> > Please change the API back to the Node.js standard of a factory export.

> Could you clarify, maybe with an example?

`app.use(morgan('dev'))`

ficristo:
I submitted a new PR with a different method in https://github.com/expressjs/morgan/pull/90

PR #90, opened 2015-11-23
Add the choice to select the color for the 'dev' format

ficristo:


ficristo:
@dougwilson Any news?

dougwilson:
Hi @ficristo, I'm sorry I haven't gotten back to yet, just been very busy :) The entire Express.js organization is part of the Node.js foundation now! This means that we'll get more eyes on this type of thing coming soon.

As far as this pull request goes, I have fixed feelings about it:

1. The idea of being able to configure the colors of the dev format is a good one, but the pull request I feel is lacking in a good approach to achieving this (like the previous PRs).
2. This PR seems to be focused on only solving this specific one thing rather than being a generic way. For example, the `devColor` option is going to be passed to every single format function, which is ripe for weird abuse. This PR is also passing it to every single token function for some reason as well.

In the meantime, there are three ways to get the grey color:

1. Just use `morgan.format()` to add a second dev format with the color you want, or use it to replace the existing `dev` format, then use that format.
2. Just provide your own function as the `format` option directly that achieves the output & color you desire.
3. Use the built-in `dev` format and set the `stream` option to a through stream that will transform the colors and pipe that to `process.stdout`.

ficristo:
> the pull request I feel is lacking in a good approach

I tried a couple of things but I wasn't really happy with any of them...

> there are three ways to get the grey color

I didn't realize it was possible, even if kind of obvious.
If possible, I think would be really nice to have something in the documentation about this (at least for me :stuck_out_tongue:).

Thank you for your time.

PR #94, opened 2016-01-13
Using npm module remote-ip to get remote ip address

guidoman:
The remote address will not be displayed correctly if e.g. the application is behind Nginx. Using NPM module "remote-ip" is a much more reliable solution IMHO.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ var debug = require('debug')('morgan')
 var deprecate = require('depd')('morgan')
 var onFinished = require('on-finished')
 var onHeaders = require('on-headers')
+var requestIp = require('request-ip')
 
 /**
  * Array of CLF month names.
@@ -455,10 +456,7 @@ function getFormatFunction(name) {
  */
 
 function getip(req) {
-  return req.ip
-    || req._remoteAddress
-    || (req.connection && req.connection.remoteAddress)
-    || undefined;
+  return requestIp.getClientIp(req)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


dougwilson:
Hi! Thank you for your pull request! Unfortunately I don't accept this type of change in this module for a few reasons:

1. We direct users to use their own strategy and simply populate `req.ip` with the IP they determined the client to have; the rest of the code is to read the IP from the socket if they did not do this.
2. The used `reqest-ip` module here is greatly insecure, in that it will sequentially read from a lot of different request headers. Unless every user who uses Morgan is fully aware of all those headers and sets up their edge load balance or their app to strip all un-trusted ones, it is just too easy for a user to just add that to their request to "forge" the IP address you see in the logs. This would make the IP entry useless.
3. It looks like the tests are failing with the changes here.

If you wish to use `request-ip` in your code, simply do the following:
```js
app.use(function (req, res, next) {
  Object.defineProperty(req, 'ip', {
    value: requestIp(req)
  })
  next()
})
app.use(morgan('combined'))
```

dougwilson:
And if the pull request https://github.com/pbojinov/request-ip/pull/12/files on that module is ever merged, then the above code would become the following:
```js
app.use(requestIp.mw({ attributeName: 'ip' }))
app.use(morgan('combined'))
```

guidoman:
OK, thanks for your detailed explanation. Now I understand how `req.ip` is supposed to work. Maybe the best solution is to have Express handle this with the "trust proxy" setting?
```
app.set('trust proxy', true)
app.use(morgan('combined'))
```

dougwilson:
Hi @guidoman , ah, sorry, I wasn't sure if you were using Express or not (this module is compatible with more than Express). Yes, if you are using Express, you can use the `trust proxy` setting to populate the `req.ip` value, if the setting works for your use-case :)

PR #98, opened 2016-03-08
Update index.js

lijason1121:
add 'locale' format by :date[locale]

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ The current date and time in UTC. The available formats are:
 
   - `clf` for the common log format (`"10/Oct/2000:13:55:36 +0000"`)
   - `iso` for the common ISO 8601 date time format (`2000-10-10T13:55:36.000Z`)
+  - `locale` for the locale time format (`3/9/2016, 13:55:36 PM`)
   - `web` for the common RFC 1123 date time format (`Tue, 10 Oct 2000 13:55:36 GMT`)
 
 If no format is given, then the default is `web`.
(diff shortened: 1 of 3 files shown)


dougwilson:
Hi @lijason1121, thanks for your pull request! I have a few comments on this so far:

1. There are no tests; please add complete tests for your feature.
2. There is no documentation; please add documentation fro your feature in the README.
3. I''m not sure what this "local" format maps to. It seems to be date/month/year hour:mins:secs +0000. What format is this? What makes this a "local" format, especially when many people consider "local" dates to be month/date/year, like myself?

lijason1121:
hi @dougwilson 
thank your for your feedback
I have completed the test and update README doc.
'locale' just for get local time. For me, I am in Beijing. So I want to log by GMT +8 time string, not UTC time:)

dougwilson:
Hi @lijason1121, it looks like the contents of the pull request are very different now :)

The date formats has always been a form on debate here, and in reality, we only provide the ones that access log analyzers can read out of the box and recommend you simply define the token for the format you want, which is the whole point of the token method:
```js

morgan.token('date', function () {
  var dateTime = new Date()
  var date = dateTime.getDate()
  var hour = dateTime.getHours()
  var mins = dateTime.getMinutes()
  var secs = dateTime.getSeconds()
  var year = dateTime.getFullYear()
  var month = dateTime.getMonth() + 1;

  return year + '-' + pad2(month) + '-' + pad2(date)
    + ' ' + pad2(hour) + ':' + pad2(mins) + ':' + pad2(secs) 
})
```

PR #103, opened 2016-04-08
kayvee request logging

nathanleiby:
Here's a prototype of node request middleware, extending `morgan`. We can decide whether to continue to extend morgan or remove all unnnecessary parts, later.

To add this middleware, currently it is invoked via `app.use(morgan('clever'))`

----

Start the express server

```
npm install
node server.js
```

Calling `curl "localhost:3000/hello/world?a=1&b=2"` results in this log being produced:

````
{"method":"GET","path":"/hello/world","params":"?a=1&b=2","response-size":"13","response-time":792955,"status-code":200}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,8 @@ var debug = require('debug')('morgan')
 var deprecate = require('depd')('morgan')
 var onFinished = require('on-finished')
 var onHeaders = require('on-headers')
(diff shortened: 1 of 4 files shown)


nathanleiby:
Sorry, didn't mean to open against upstream. thanks!

PR #105, opened 2016-05-03
Adding comment to README about statement ordering

apjanes:
Much trouble was had before we learned that the: `app.use(morgan(...));` statement needs to occur straight after the creation of the app before other `use` calls.  Notes in the README file would have helped a lot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,6 @@ var express = require('express')
 var morgan = require('morgan')
 
 var app = express()
-
 app.use(morgan('combined'))
 
 app.get('/', function (req, res) {
@@ -239,6 +238,9 @@ var app = express()
 var accessLogStream = fs.createWriteStream(__dirname + '/access.log', {flags: 'a'})
 
 // setup the logger
+// NOTE: the use of morgan needs to be done immediately after the creation
+//       of express, before any other use calls are made.  If it is not,
+//       log files may be created but may not be written to.
 app.use(morgan('combined', {stream: accessLogStream}))
 
 app.get('/', function (req, res) {
@@ -273,6 +275,9 @@ var accessLogStream = FileStreamRotator.getStream({
 })
 
 // setup the logger
+// NOTE: the use of morgan needs to be done immediately after the creation
+//       of express, before any other use calls are made.  If it is not,
+//       log files may be created but may not be written to.
 app.use(morgan('combined', {stream: accessLogStream}))
 
 app.get('/', function (req, res) {
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# morgan
+n
 
 [![NPM Version][npm-image]][npm-url]
 [![NPM Downloads][downloads-image]][downloads-url]
@@ -195,6 +195,21 @@ app.get('/', function (req, res) {
   res.send('hello, world!')
 })
 ```
+**note:**
+when using morgan with express, it is important that morgan is used by the express app before any routing is configured.  For example, this will **not** work:
+```javascript
+var express = require('express')
+var morgan = require('morgan')
+
+var app = express()
+
+app.get('/', function (req, res) {
+  res.send('hello, world!')
+})
+app.use(morgan('combined'))
+```
+This is because the usage of morgan is done after the configuration of the route.
+This also applies to the configuration of static resources.
 
 ### vanilla http server
 
@@ -238,9 +253,6 @@ var app = express()
 var accessLogStream = fs.createWriteStream(__dirname + '/access.log', {flags: 'a'})
 
 // setup the logger
-// NOTE: the use of morgan needs to be done immediately after the creation
-//       of express, before any other use calls are made.  If it is not,
-//       log files may be created but may not be written to.
 app.use(morgan('combined', {stream: accessLogStream}))
 
 app.get('/', function (req, res) {
@@ -327,3 +339,4 @@ function assignId(req, res, next) {
 [downloads-url]: https://npmjs.org/package/morgan
 [gratipay-image]: https://img.shields.io/gratipay/dougwilson.svg
 [gratipay-url]: https://www.gratipay.com/dougwilson/
+


dougwilson:
Thanks! I do have some issues with your changes though:

1. You make it sound like a user "needs to" put this as the first middleware. This is absolutely not true. Sure, it may be nice for many use-cases, but not all. For example, many people do not want to log the access to all the favicon requests, so put it below that middleware.
2. You wrote if not, log files may be created but not written to. What? Is this a bug? Can you elaborate to me what is going on here? I would rather fix bugs than just write them down :)

Overall, I think the changes need to be actual documentation, not hidden in comments in the examples. The documentation should describe how this module is working so people can actually understand where they need to app.use it, not just be told a one-size-does-not-fit-all statement.

apjanes:
> "You make it sound like a user "needs to" put this as the first middleware. This is absolutely not true."
 > "You wrote if not, log files may be created but not written to. What? Is this a bug? Can you elaborate..."

Yes and yes :)  In our experience it DOES matter what order initialisation occurs and, if this is not the design, it may be a bug (and, I agree, should be fixed rather than documented :) )

When our `app.js` file is written like this:

```javascript
//  ... snip
var app = express();

// MORGAN SETUP
var logDirectory = __dirname + '/log';
var FileStreamRotator = require('file-stream-rotator');
fs.existsSync(logDirectory) || fs.mkdirSync(logDirectory)
var accessLogStream = FileStreamRotator.getStream({
  date_format: 'YYYYMMDD',
  filename: logDirectory + '/access-%DATE%.log',
  frequency: 'daily',
  verbose: true
});
app.use(logger('combined', {stream: accessLogStream}));
// END MORGAN SETUP

app.set('views', path.join(__dirname, 'views'));
app.set('view engine', 'hbs');
app.use(bodyParser.json());
app.use(bodyParser.urlencoded({ extended: false }));
app.use(cookieParser());
app.use(express.static(path.join(__dirname, 'public')));
app.use('/', routes);
app.use(function(req, res, next) {
  var err = new Error('Not Found');
  err.status = 404;
  next(err);
});
// snip ...
```
everything works as expected and we get data in `access-20160503.log` like this:

```
::1 - - [03/May/2016:15:09:08 +0000] "GET / HTTP/1.1" 304 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.86 Safari/537.36"
::1 - - [03/May/2016:15:09:08 +0000] "GET /images/main-logo.svg HTTP/1.1" 304 - "http://localhost:3000/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.86 Safari/537.36"
::1 - - [03/May/2016:15:09:08 +0000] "GET /stylesheets/main.css HTTP/1.1" 404 1668 "http://localhost:3000/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.86 Safari/537.36"
::1 - - [03/May/2016:15:09:08 +0000] "GET /images/footer-logo.svg HTTP/1.1" 304 - "http://localhost:3000/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.86 Safari/537.36"
```
However, when we change our `app.js` to this:
```javascript
//  ... snip
var app = express();
app.set('views', path.join(__dirname, 'views'));
app.set('view engine', 'hbs');
app.use(bodyParser.json());
app.use(bodyParser.urlencoded({ extended: false }));
app.use(cookieParser());
app.use(express.static(path.join(__dirname, 'public')));
app.use('/', routes);
app.use(function(req, res, next) {
  var err = new Error('Not Found');
  err.status = 404;
  next(err);
});

// MORGAN SETUP
var logDirectory = __dirname + '/log';
var FileStreamRotator = require('file-stream-rotator');
fs.existsSync(logDirectory) || fs.mkdirSync(logDirectory)
var accessLogStream = FileStreamRotator.getStream({
  date_format: 'YYYYMMDD',
  filename: logDirectory + '/access-%DATE%.log',
  frequency: 'daily',
  verbose: true
});
app.use(logger('combined', {stream: accessLogStream}));
// END MORGAN SETUP

// snip ...
```
the `access-20160503.log` file is still created but no data is logged.  Am I missing something?  Is this expected behaviour?

dougwilson:
Right, order does matter, but it is not a straight forward as your PR. Your pull request is actually misinformation, bordering on just plain wrong. I understand you were frustrated on time spent debugging, but that does not make adding wrong & misleading comments to examples here acceptable.

Overall, I think the changes need to be actual documentation, not hidden in comments in the examples. The documentation should describe how this module is working so people can actually understand where they need to app.use it, not just be told a one-size-does-not-fit-all statement.

Can you please update your pull request to actually describe how the module is working and help guide people to understand where they should be ordering this middleware. In my apps I almost never have this module as the first middleware, because that rarely makes sense for large apps. We need to actually document the necessary information for people to understand where to place it, not just say put it as the top. Does that make sense?

Once you document & understand how this module is working, you'll actually pretty easily understand why your examples have the behavior they do. It doesn't look like a bug, especially since this module is not creating the access log file, your code is.

dougwilson:
In addition, the documentation should target users who understand how Express routing in general. If Express routing documentation does not explain why your second example is just completely wrong from an Express standpoint, then you may also want to file an issue or make a pull request to the Express documentation. The placement of Morgan at the bottom would never work with any middleware, so your issue is not even with this module, perhaps with the understanding of how Express works.

apjanes:
Okay.  Several things here:

1. I sense irritation bordering on crossness in your tone.  If I have caused this, I apologise.  While I have been developing for a long time, I new to contributing to a public git project.  Perhaps I should not have done this via a pull request.

1. The call to `logger` is actually a call to morgan.  I realised after writing my comments that we are, unhelpfully, using `var logger = require('morgan');` rather than using `var morgan` as per the examples.  Sorry if this caused confusion.  I will edit my comments to use the name morgan instead.

1. "I understand you were frustrated on time spent debugging, but that does not make adding wrong & misleading comments to examples here acceptable": absolutely, it is never my intention to do so.  My comments were describing what worked for us without a complete understanding as to why, once again, apologies

1. Further investigation has let me to revise my understanding.  Unless morgan is initialised _before_ what is being logged (either through route creation or the creation of a static location) the data for that resource is not logged.  For example, if I initialize before `app.use('/', routes);` but before `app.use(express.static(...));` then the requests to the routes are logged but any requests for static resources aren't.

1. While we are using the more complex file rotation logger, the same symptoms exist when using the basic `app.use(morgan('combined'));`.  If this line occurs after the registration of routers/static paths etc, they don't get logged.  While there may be a very valid reason for this, it does not appear to be documented and I believe it would help developers using morgan avoid wasting time if it were.  This was my aim with the pull request even though it appears to be poorly executed.

1. If you are unhappy with the wording and location of the documentation, would you please suggest an alternative wording and/or location?

apjanes:
> The placement of Morgan at the bottom would never work with any middleware, so your issue is not even with this module, perhaps with the understanding of how Express works.

Perhaps it is.  However, I would wager that there are other developers who are also missing some understanding and, providing some details about where morgan should be located in the express initialisation process and why would be helpful.  If you do not wish to do this or would prefer to do this yourself with greater accuracy, I will delete this pull request.  

If, instead, you would like me to revise my changes to make them more accurate or located them elsewhere, I am happy to do that.

dougwilson:
I'm sorry if my tone came across weird. I am trying to write all this on a phone and it is difficult to really respond. Your points are too long for me to reply on a phone. Is it OK if we postpone this discussion for a week or so when I will be back at a computer so I can fully read and respond to all points?

I would like it if you could adjust the PR, which is why I have been trying to provide feedback, sorry. I can always do it myself but that may not be for months, so if you can provide the time and effort, that would be amazing.

I'll come back to this PR in about a week and a half when I get back from vacation. Thanks for the understanding.

apjanes:
No problem about the tone.  I didn't realise you were on a mobile.  I've updated the README to remove the comment in the example code and instead, in the express/connect section I've added the note:

when using morgan with express, it is important that morgan is used by the express app before any routing is configured.

I have also added an example of what NOT to do.  I hope this is better.  If you feel it still needs adjustment or I am still not accurate, please say.

dougwilson:
Thank you, but after Express.js discussion, we are not going to accept this. We have in the past tried to show examples of what not to do, but people literally just don't read the words and simply copy-and-paste examples, and they have always been too much trouble to show any code that is a "do not do this example".

In addition, the PR removes the title of this module and the misleading comment is still there (though you removed it from one of the examples, but just not the other). Since there is nothing in the pull request that looks good, there is nothing to accept.

You are welcome to create a new pull request if you like. What we are looking for is the following: adding a description of how the module is actually working, such that people can reasonable about where the placement needs to be for their needs, not just giving bland information and not understand what they are doing, rather than [running into the money & ladder issue](http://www.wisdompills.com/2014/05/28/the-famous-social-experiment-5-monkeys-a-ladder/).

PR #107, opened 2016-05-21
Color Status Token

frankydp:
Just an additional default token for colorized status codes.  Not sure if it would be wanted or useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,15 @@ morgan.token('type', function(req, res){ return req.headers['content-type']; })
 
 Calling `morgan.token()` using the same name as an existing token will overwrite that token definition.
 
+##### :colorstatus
+
+The colorized :status of the response.  Inserts the status code with colors applied coordinating to their group.
+ - `2XX` - Green
+ - `3XX` - Cyan
+ - `4XX` - Yellow
+ - `5XX` - Red
+
+
 ##### :date[format]
 
 The current date and time in UTC. The available formats are:
@@ -158,10 +167,6 @@ The status code of the response.
 If the request/response cycle completes before a response was sent to the
 client (for example, the TCP socket closed prematurely by a client aborting
 the request), then the status will be empty (displayed as `"-"` in the log).
-
-##### :colorstatus
-
-The colorized :status of the response.
 
 ##### :url
 
--- a/README.md
+++ b/README.md
@@ -159,6 +159,10 @@ If the request/response cycle completes before a response was sent to the
 client (for example, the TCP socket closed prematurely by a client aborting
 the request), then the status will be empty (displayed as `"-"` in the log).
 
+##### :colorstatus
+
+The colorized :status of the response.
+
 ##### :url
 
 The URL of the request. This will use `req.originalUrl` if exists, otherwise `req.url`.
--- a/test/morgan.js
+++ b/test/morgan.js
@@ -1108,7 +1108,7 @@ describe('morgan()', function () {
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
           var masked = line.replace(/\x1b\[(\d+)m/g, '_color_$1_')
-          assert.equal(masked.substr(0, 28), 'GET / _color_0_102 _color_0_')
+          assert.equal(masked.substr(0, 38), '_color_0_ GET / _color_0_102 _color_0_')
           done()
         })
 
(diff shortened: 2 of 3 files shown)


dougwilson:
Thanks! The concept seems fine, just a few issues:

1. The changes do not even compile. All tests are red due to the inability to now even `require()` the script. Please take a look into this issue.
2. There are no tests for the added token, please add complete tests for app possible branches of the token :)
3. There is no documentation added to the README for the token, so that would be sweet to get added as well :)

In addition to this, this PR should also DRY up the `dev` format as well. Probably the code for this new token should be the exact same color code from the `dev` format instead of some other code, and the format can just use this token and then get smashed down into just being a string like the other formats are.

frankydp:
Thanks for the feedback.  I will get the details hashed out and resub.

dougwilson:
Cool :) Just update this pull request with the new changes and ping me when they are included and I'll take another look :)

frankydp:
@dougwilson Dried up the dev compile to the string format.  Added test for new token, and corrected test for dev to account for the resulting changes to responses, left in the color test in dev even though they were a little redundant.  Added colorstatus token to docs.  Also, changed the color code to be more inline with the original implementation from dev.  Sorry if there was any build spam.

README.md
```diff
@@ -159,6 +159,10 @@ If the request/response cycle completes before a response was sent to the
 client (for example, the TCP socket closed prematurely by a client aborting
 the request), then the status will be empty (displayed as `"-"` in the log).
 
+##### :colorstatus
+
+The colorized :status of the response.
```
dougwilson: Please add documentation about which statuses get which colors.

README.md
```diff
@@ -159,6 +159,10 @@ If the request/response cycle completes before a response was sent to the
 client (for example, the TCP socket closed prematurely by a client aborting
 the request), then the status will be empty (displayed as `"-"` in the log).
 
+##### :colorstatus
```
dougwilson: Currently they are just listed in alphabetical order, if you can move that up where "c" would go.

test/morgan.js
```diff
@@ -1067,7 +1219,7 @@ describe('morgan()', function () {
           var cb = after(2, function (err, res, line) {
             if (err) return done(err)
             var masked = line.replace(/\x1b\[(\d+)m/g, '_color_$1_')
-            assert.equal(masked, '_color_0_GET / _color_0_- _color_0_- ms - -_color_0_')
```
dougwilson: Please put back the leading color resets. It was added because there were cases when the string would print, but the previous thing did not reset it's color, so the beginning would end up being a random color.

frankydp: Removed them due to the change in the actual responses when using 'dev' after changes.  Since it was dried up the response no longer is wrapped in the color resets, because the default tokens do not have them.  

So, a couple options, I can rollback the the changes to dev, or add a color reset to all functions via the compile(which may break all current response test), or leave it as is, due to the now inserted token having its own resets on front and back for all returned values.  

Appreciate any advice.

dougwilson: Hmm. Well, the color resets is a necessary feature we must have in the `dev` token, so adding that back somehow isn't even a question in my mind.

Adding the color reset in the other tokens absolutely cannot happen, either, as all those other formats are being written out to log files and the like and this will be very heavily breaking, since the resets are actual characters that are going to end up in all those log files all of a sudden.

I'm not sure what a good solution is right now, because I'd have to think about it. If you can proposal any additional ideas, I'd be happy to read them!

arikalfus:
The last comment on this was on May 28. Is this still on track to being implemented @frankydp? Seems like a revert of a removed test is still pending.

frankydp:
@arikalfus Revert was committed, with a clear added to accommodate, back on May 28th.  Pull seems to have conflicts now though.

dougwilson:
Hi @frankydp , sorry, until very recently GitHub did not send any notice when PRs where updated, so since your last update you never posted a comment, I didn't realize you pushed anything else, very sorry for the misunderstanding!

Looking over the PR, there are still a few issues to be resolved:

1. Every line in the `:dev` format seems to now be starting with a space character where before the first letter of the method was the first character. You should probably remove that space character.
2. The documentation for `:colorstatus` is confusing. It sounds like it's just the colorized `:status` token, but if I override that token to do something different, `:colorstatus` does not seem to be affected.
3. It doesn't seem like you reverted those tests that I asked earlier, as they are still showing as altered in the current diff (one line was changed and another line was removed). Can you put them back to how they were, please?

dougwilson:
Closing the PR since I never heard back and there are conflicts that prevent merging.

PR #111, opened 2016-06-15
Adding 'local' format to date token

joansanfeliu:
A small patch to add a local format to the date token. It uses a format similar to clf using the current local time instead of UTC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ The current date and time in UTC. The available formats are:
   - `clf` for the common log format (`"10/Oct/2000:13:55:36 +0000"`)
   - `iso` for the common ISO 8601 date time format (`2000-10-10T13:55:36.000Z`)
   - `web` for the common RFC 1123 date time format (`Tue, 10 Oct 2000 13:55:36 GMT`)
+  - `local` for the local string format (`"10/Oct/2000:15:55:36 UTC+2"`)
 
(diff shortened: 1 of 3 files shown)


dougwilson:
Hi, and thanks for the pull request! Unfortunately we are not going to accept this, and there are many closed conversations with background already.

You can easily add custom tokens to your own Morgan instance, which you can do to add the local time format you desire.

Also, it looks like this format is kind of broken for time zones that at not offset from UTC by full hours, for example India is offset by an additional 30 minutes and Nepal is offset by an additional 45 minutes.

Also, I know you said it uses a format similar to CLF, but it may have been more useful if it actually was the CLF instead of your own twist on how to represent the time zone.

joansanfeliu:
Hi, 
I see what you mean and I agree. Adding a custom token will suit my needs just fine. 
Thanks!

dougwilson:
I threw together a `:localdate` token that outputs the same format as this PR (though the time zone is slightly different) using the popular Moment.js module:
```js
var moment = require('moment') // npm i moment@2.13.0
var morgan = require('morgan') // npm i morgan@1.7.0


morgan.token('localdate', function getLocalDateToken (req, res) {
  return moment().format('DD/MMM/YYYY:HH:mm:ss UTCZZ')
})
```

PR #124, opened 2016-10-06
Doc custom

bmeck:
Superseding : https://github.com/expressjs/morgan/pull/123
Dropped for: #125 #127

Documents custom tokens and formatters a bit more/makes the API public knowledge in order to keep it stable in future. Also adds a few tests. Used *Each in tests because `after` doesn't work as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,11 +356,11 @@ instead of a simple string.
 var express = require('express')
 var morgan = require('morgan')
 
-var combined = morgan.compile('[:date[clf]] :method :url :res[content-length]')
+var formatter = morgan.compile('[:date[clf]] :method :url :res[content-length]')
 var JSONOutput = function (morgan, req, res) {
   return JSON.stringify({
     status: morgan.status(req,res),
-    default: combined(morgan, req, res),
+    default: formatter(morgan, req, res),
   })
 }
 
--- a/README.md
+++ b/README.md
@@ -327,20 +327,20 @@ function assignId (req, res, next) {
 
 #### arguments to custom token formats
 
-Example of a custom token format that uses an argument. It will echo the
-argument passed to the token.
+Example of a custom token format that uses an argument. It will output the
+header related to the argument passed to the token.
 
 ```js
 var express = require('express')
 var morgan = require('morgan')
 
-morgan.token('echo', function getId (req, res, arg) {
-  return arg
+morgan.token('header', function getId (req, res, arg) {
+  return String(req.headers[arg])
 })
 
 var app = express()
 
-app.use(morgan(':echo[hello world!] :response-time'))
+app.use(morgan(':header[host] :response-time'))
 
 app.get('/', function (req, res) {
   res.send('hello, world!')
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ Create a new morgan logger middleware function using the given `format` and `opt
 The `format` argument may be a string of a predefined name (see below for the names),
 a string of a format string, or a function that will produce a log entry.
 
+For the syntax of strings and functions see [morgan.compile](#morgancompileformat).
+
 #### Options
 
 Morgan accepts these properties in the options object.
@@ -103,6 +105,14 @@ To define a token, simply invoke `morgan.token()` with the name and a callback f
 morgan.token('type', function (req, res) { return req.headers['content-type'] })
 ```
 
+Tokens can accept a string argument passed in from `[]` brackets by specifying
+a third argument `arg`:
+```js
+morgan.token('type', function (req, res, arg) { return arg })
+```
+
+Falsey values returned from this function will be replaced with `-`.
+
 Calling `morgan.token()` using the same name as an existing token will overwrite that token definition.
 
 ##### :date[format]
@@ -177,6 +187,11 @@ be passed using `[]`, for example: `:token-name[pretty]` would pass the string
 
 Normally formats are defined using `morgan.format(name, format)`, but for certain
 advanced uses, this compile function is directly available.
+
+The function returned takes arguments `tokens` , `req`, and `res` where `tokens`
+refers to `morgan` itself. If a log should be skipped the function will return
+`null`. The function returned or a custom function can be passed directly to 
+morgan using `morgan(myFn)`.
 
 ## Examples
 
@@ -308,6 +323,54 @@ function assignId (req, res, next) {
   req.id = uuid.v4()
   next()
 }
+```
+
+#### arguments to custom token formats
+
+Example of a custom token format that uses an argument. It will echo the
+argument passed to the token.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+morgan.token('echo', function getId (req, res, arg) {
+  return arg
+})
+
+var app = express()
+
+app.use(morgan(':echo[hello world!] :response-time'))
+
+app.get('/', function (req, res) {
+  res.send('hello, world!')
+})
+```
+
+### use custom format function
+
+Sample app that will use a custom formatting function. This will print JSON
+instead of a simple string.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+var combined = morgan.compile('[:date[clf]] :method :url :res[content-length]')
+var JSONOutput = function (morgan, req, res) {
+  return JSON.stringify({
+    status: morgan.status(req,res),
+    default: combined(morgan, req, res),
+  })
+}
+
+var app = express()
+
+app.use(morgan(JSONOutput))
+
+app.get('/', function (req, res) {
+  res.send('hello, world!')
+})
 ```
 
 ## License
--- a/test/morgan.js
+++ b/test/morgan.js
@@ -825,47 +825,61 @@ describe('morgan()', function () {
     })
 
     describe('a custom token function provided to .token(name,fn)', function () {
-      function token (req, res, arg) {
-        return arg && url.parse(req.url)[arg]
-      }
-      var urlToken = morgan.url
-      beforeEach(function (done) {
-        // test clobbering
+      it('should overwrite existing', function (done) {
+        function token (req, res, arg) {
+          return arg && url.parse(req.url)[arg]
+        }
+        var urlToken = morgan.url
         morgan.token('url', token)
-        done();
-      })
-      afterEach(function (done) {
+        assert.equal(morgan.url, token)
         morgan.token('url', urlToken)
         done()
       })
 
-      it('should overwrite existing', function (done) {
-        assert.equal(morgan.url, token)
-        done()
-      })
+      describe('should use the string `-` if return value of the token function is falsey', function () {
+        function test (v, done) {
+          morgan.token('ret', function ret (req, res, arg) {
+            return arg
+          })
+          var cb = after(2, function (err, res, line) {
+            if (err) return done(err)
+            assert.equal(line, '-')
+            done()
+          })
 
-      it('should use the string `-` if return value of the token function is falsey', function (done) {
-        var cb = after(2, function (err, res, line) {
-          if (err) return done(err)
-          assert.equal(line, '-')
-          done()
+          var stream = createLineStream(function (line) {
+            cb(null, null, line)
+          })
+
+          var server = createServer(':ret', { stream: stream }, function (req, res, next) {
+            next()
+          })
+
+          request(server)
+          .get('/')
+          .expect(200, cb)
+        }
+        [
+          0,
+          NaN,
+          false,
+          '',
+          null,
+          undefined
+        ].forEach(function (v) {
+          it('for ' + (String(v) || JSON.stringify(v)), function (done) {
+            test(v, done)
+          })
         })
-
-        var stream = createLineStream(function (line) {
-          cb(null, null, line)
-        })
-
-        var server = createServer(':url', { stream: stream }, function (req, res, next) {
-          next()
-        })
-
-        request(server)
-        .get('/')
-        .expect(200, cb)
       })
 
       it('should not see empty brackets as an argument value', function (done) {
-        var cb = after(2, function (err, res, line) {
+        morgan.token('checkempty', function ret (req, res, arg) {
+          assert.equal(arg, undefined)
+          assert.equal(arguments.length, 2)
+          cb(null)
+        })
+        var cb = after(3, function (err, res, line) {
           if (err) return done(err)
           assert.equal(line, '-[]')
           done()
@@ -875,7 +889,7 @@ describe('morgan()', function () {
           cb(null, null, line)
         })
 
-        var server = createServer(':url[]', { stream: stream }, function (req, res, next) {
+        var server = createServer(':checkempty[]', { stream: stream }, function (req, res, next) {
           next()
         })
 
@@ -885,22 +899,24 @@ describe('morgan()', function () {
       })
 
       it('should use the result', function (done) {
+        morgan.token('urlpart', function ret (req, res, arg) {
+          return url.parse(req.url)[arg]
+        })
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
           assert.equal(line, 'foo=bar')
           done()
         })
 
-
         var stream = createLineStream(function (line) {
           cb(null, null, line)
         })
 
-        var server = createServer(':url[query]', { stream: stream }, function (req, res, next) {
+        var server = createServer(':urlpart[query]', { stream: stream }, function (req, res, next) {
           next()
         })
 
-        var test = request(server).post('/test').query('foo=bar').expect(200, cb)
+        request(server).post('/test').query('foo=bar').expect(200, cb)
       })
     })
   })
--- a/test/morgan.js
+++ b/test/morgan.js
@@ -823,102 +823,6 @@ describe('morgan()', function () {
         test.write('0')
       })
     })
-
-    describe('a custom token function provided to .token(name,fn)', function () {
-      it('should overwrite existing', function (done) {
-        function token (req, res, arg) {
-          return arg && url.parse(req.url)[arg]
-        }
-        var urlToken = morgan.url
-        morgan.token('url', token)
-        assert.equal(morgan.url, token)
-        morgan.token('url', urlToken)
-        done()
-      })
-
-      describe('should use the string `-` if return value of the token function is falsey', function () {
-        function test (v, done) {
-          morgan.token('ret', function ret (req, res, arg) {
-            return arg
-          })
-          var cb = after(2, function (err, res, line) {
-            if (err) return done(err)
-            assert.equal(line, '-')
-            done()
-          })
-
-          var stream = createLineStream(function (line) {
-            cb(null, null, line)
-          })
-
-          var server = createServer(':ret', { stream: stream }, function (req, res, next) {
-            next()
-          })
-
-          request(server)
-          .get('/')
-          .expect(200, cb)
-        }
-        [
-          0,
-          NaN,
-          false,
-          '',
-          null,
-          undefined
-        ].forEach(function (v) {
-          it('for ' + (String(v) || JSON.stringify(v)), function (done) {
-            test(v, done)
-          })
-        })
-      })
-
-      it('should not see empty brackets as an argument value', function (done) {
-        morgan.token('checkempty', function ret (req, res, arg) {
-          assert.equal(arg, undefined)
-          assert.equal(arguments.length, 2)
-          cb(null)
-        })
-        var cb = after(3, function (err, res, line) {
-          if (err) return done(err)
-          assert.equal(line, '-[]')
-          done()
-        })
-
-        var stream = createLineStream(function (line) {
-          cb(null, null, line)
-        })
-
-        var server = createServer(':checkempty[]', { stream: stream }, function (req, res, next) {
-          next()
-        })
-
-        request(server)
-        .get('/')
-        .expect(200, cb)
-      })
-
-      it('should use the result', function (done) {
-        morgan.token('urlpart', function ret (req, res, arg) {
-          return url.parse(req.url)[arg]
-        })
-        var cb = after(2, function (err, res, line) {
-          if (err) return done(err)
-          assert.equal(line, 'foo=bar')
-          done()
-        })
-
-        var stream = createLineStream(function (line) {
-          cb(null, null, line)
-        })
-
-        var server = createServer(':urlpart[query]', { stream: stream }, function (req, res, next) {
-          next()
-        })
-
-        request(server).post('/test').query('foo=bar').expect(200, cb)
-      })
-    })
   })
 
   describe('formats', function () {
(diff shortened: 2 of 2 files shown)


dougwilson:
Thanks, @bmeck . Looks like there is a lot going on in this PR than just a docs change, so will need some time to look through it.

README.md
```diff
@@ -178,6 +188,11 @@ be passed using `[]`, for example: `:token-name[pretty]` would pass the string
 Normally formats are defined using `morgan.format(name, format)`, but for certain
 advanced uses, this compile function is directly available.
 
+The function returned takes arguments `tokens` , `req`, and `res` where `tokens`
+refers to `morgan` itself. If a log should be skipped the function will return
+`null`. The function returned or a custom function can be passed directly to
```
dougwilson: I think adding it somewhere else is a lot clearer; adding the information for the function that you pass into the first argument of morgan way down here seems too far out of the way to me.

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
...
+
+Example of a custom token format that uses an argument. It will echo the
+argument passed to the token.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+morgan.token('echo', function getId (req, res, arg) {
+  return arg
```
dougwilson: The examples should be something someone will actually do in the real world. Please re-work the example to something a user would actually do.

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
...
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+var combined = morgan.compile('[:date[clf]] :method :url :res[content-length]')
+var JSONOutput = function (morgan, req, res) {
+  return JSON.stringify({
+    status: morgan.status(req,res),
+    default: combined(morgan, req, res),
```
dougwilson: I'm confused; would someone do this in a real app? Not sure why someone would do this if you can elaborate on that.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
         test.write('0')
       })
     })
+
+    describe('a custom', function () {
```
dougwilson: I'm not sure I understand the description. This right now reads in the unit tests "morgan() tokens a custom" ??

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
+      beforeEach(function () {
+        url = morgan.url
+        morgan.token('url', token)
+      })
+      afterEach(function () {
+        morgan.url = url
+      })
+
+      it('should overwrite existing', function (done) {
+        assert.equal(morgan.url, token)
```
dougwilson: Hm. I know that is ends up as a property on `morgan`, but I'm not sure if I want to really promise that functionality... Will have to think on this.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
         test.write('0')
       })
     })
+
+    describe('a custom', function () {
+      function token (req, res, arg) {
+        return arg
```
dougwilson: Can the test be something more real? For one thing, none of the tests here ever actually validate the `req` and `res` are actually those types of objects since this function is so fake. If it was real, then you can read real properties off real requests/responses, which i.m.o. are better real tests that would detect a variety of issues.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
+      }
+      var url
+      beforeEach(function () {
+        url = morgan.url
+        morgan.token('url', token)
+      })
+      afterEach(function () {
+        morgan.url = url
```
dougwilson: I don't think this should be something we really guarantee; people need to be using the `.token` function.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
     })
+
+    describe('a custom', function () {
+      function token (req, res, arg) {
+        return arg
+      }
+      var url
+      beforeEach(function () {
+        url = morgan.url
+        morgan.token('url', token)
```
dougwilson: Why even bother overwriting an existing token? Are you testing the ability to overwrite already-existing tokens? That seems like it should be a different test, as these tests seem to be trying to test a custom token function, in which case, defining a token with a new name would be a lot clearer in the below tests.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
+      afterEach(function () {
+        morgan.url = url
+      })
+
+      it('should overwrite existing', function (done) {
+        assert.equal(morgan.url, token)
+        done()
+      })
+
+      it('should use - if result is falsey', function (done) {
```
dougwilson: I think the test description is weird here (for reference, it is: "morgan() tokens a custom should use - if result is falsey"; result of what? use it where?). Assuming the description above this is changed, this it should maybe something along the lines of "should use "-" as the token's value"--probably need to write somewhere that that is when the token returns "undefined" if you want to test falsy, you should really add more tests that guarantee that, like probably one for `false`, `null`, `0`, and `''`, which are all very interesting.

Since the interface is said to return a string, you may be able to skip the `false` and `0` I guess.

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
+        var server = createServer(':url', { stream: stream }, function (req, res, next) {
+          next()
+        })
+
+        request(server)
+        .get('/')
+        .expect(200, cb)
+      })
+
+      it('should not see empty brackets', function (done) {
```
dougwilson: What should not see the empty brackets? I feel like this test doesn't belong in this describe at all; it seems to be testing the parsing of the format string itself, not the functioning of the token itself?

test/morgan.js
```diff
@@ -822,6 +822,79 @@ describe('morgan()', function () {
...
+      })
+
+      it('should use the result', function (done) {
+        var stream = createLineStream(function (line) {
+          assert.equal(line, 'HIDDEN')
+          server.close(done)
+        })
+
+        var server = createServer(':url[HIDDEN]', { stream: stream }, function () {
+          test.abort()
```
dougwilson: I don't understand the propose of the request abort here. Can you elaborate on that?

README.md
```diff
@@ -103,6 +105,14 @@ To define a token, simply invoke `morgan.token()` with the name and a callback f
 morgan.token('type', function (req, res) { return req.headers['content-type'] })
 ```
 
+Tokens can accept a string argument passed in from `[]` brackets by specifying
+a third argument `arg`:
+```js
+morgan.token('type', function (req, res, arg) { return arg })
```
dougwilson: Use a more realistic example, please :)

README.md
bmeck: we do it currently so that we can use JSON metadata queries on the logs.

test/morgan.js
bmeck: I tried to match the `a function` in the same file, can rename

test/morgan.js
bmeck: the other tests for tokens also don't confirm this, unclear on intent of adding those tests to this specific one.

test/morgan.js
bmeck: because thats what the docs say happens.

test/morgan.js
bmeck: we could remove the clobbering behavior/note that it is a property from the docs?

test/morgan.js
bmeck: would need to change the API contract of the arguments of `morgan()` but seems doable.

test/morgan.js
bmeck: the interface returns an `any` and uses `returnValue || '-'` in https://github.com/expressjs/morgan/blob/1533cdcc512f4daf20a93cb0ab3b18d6563a897f/index.js#L387 it is not only a string return value, returning `""` also invokes this behavior.

test/morgan.js
bmeck: the token function doesn't see empty backets, it gets no argument, and the `[]` is just added to the output string.

README.md
dougwilson: I can put some thought into it over the weekend if you're not able to think of anything, no problem :)

test/morgan.js
bmeck: it was copy paste, unclear why its in other tests, so I left it.

test/morgan.js
dougwilson: Sure, but "a" is an article and "function" in a noun, so it makes sense. I assume that you're not referring to "custom" the noun here, though you're using it as a noun. I guess it sounds weird since the string doesn't end in the noun you are trying to mean.

README.md
bmeck: should we move the description of the string/id to the top as well?

test/morgan.js
dougwilson: I'm talking about what we promise from an API. Please change to describe the promises we're trying to make here.

test/morgan.js
dougwilson: Then add all the other tests.

test/morgan.js
dougwilson: But that is a test of the format parsing, not the token itself; doesn't belong in this describe.

README.md
dougwilson: I guess

dougwilson: I still think it feels like you just stuck this in a random spot (when reading the docs--why do I need to read the `.compile()` function to understand what to pass into `format`? The current docs direct to https://github.com/expressjs/morgan#predefined-formats not the compile() function.

README.md
dougwilson: Is it possible to use another variable than `combined` since it's not the combined format? Also, can you use a named function instead of an anonymous function? Or you can use an anonymous function directly in the `morgan()` call.

test/morgan.js
bmeck: what is a condoned way to check this...

test/morgan.js
bmeck: i am writing tests against the docs, unclear why I need different ones.

README.md
bmeck: fixed, example moving to `header[...]`

README.md
bmeck: moved to `header[...]`

README.md
bmeck: changing to `formatter`

test/morgan.js
bmeck: changed to `a custom token function provided to .token(name,fn)`

test/morgan.js
bmeck: done

test/morgan.js
bmeck: moved the custom (not predefined) token tests to its own top level describe

README.md
dougwilson: Can you just make different PRs? Perhaps at least separate the tests additions from the docs additions? I'm losing track of what is happening and what is changing without needing to try and re-read through everything over and over, and there is a lot of things.

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+      .expect(200, cb)
+    }
+    [
+      0,
+      NaN,
+      false,
+      '',
+      null,
+      undefined
+    ].forEach(function (v) {
```
dougwilson: The pattern I would like to see in the tests is they should not be DRY; I would rather see a separate `it` for every one of these.

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+
+    var stream = createLineStream(function (line) {
+      cb(null, null, line)
+    })
+
+    var server = createServer(':urlpart[query]', { stream: stream }, function (req, res, next) {
+      next()
+    })
+
+    request(server).post('/test').query('foo=bar').expect(200, cb)
```
dougwilson: Usually each of these method calls we put on a new line.

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+      if (err) return done(err)
+      assert.equal(line, 'foo=bar')
+      done()
+    })
+
+    var stream = createLineStream(function (line) {
+      cb(null, null, line)
+    })
+
+    var server = createServer(':urlpart[query]', { stream: stream }, function (req, res, next) {
```
dougwilson: It doesn't seem necessary to pas the function if it does nothing.

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+    var server = createServer(':checkempty[]', { stream: stream }, function (req, res, next) {
+      next()
+    })
+
+    request(server)
+    .get('/')
+    .expect(200, cb)
+  })
+
+  it('should use the result', function (done) {
```
dougwilson: This test says "morgan.token(name, function) should use the result". I'm not certain what that means. What should use the result? The result of what?

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+      it('for ' + (String(v) || JSON.stringify(v)), function (done) {
+        test(v, done)
+      })
+    })
+  })
+
+  it('should not see empty brackets as an argument value', function (done) {
+    morgan.token('checkempty', function ret (req, res, arg) {
+      assert.equal(arg, undefined)
+      assert.equal(arguments.length, 2)
```
dougwilson: I don't think it's generally a good practice that we should be guaranteeing the `arguments.length`.

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+      null,
+      undefined
+    ].forEach(function (v) {
+      it('for ' + (String(v) || JSON.stringify(v)), function (done) {
+        test(v, done)
+      })
+    })
+  })
+
+  it('should not see empty brackets as an argument value', function (done) {
```
dougwilson: From the earlier comment, this test is not testing tokens; it's testing the parsing of the format string. i.m.o. the ignoring of the `[]` can really be considered a bug, not a feature. Perhaps we should be it rather than engrain the bug into the design?

test/morgan.js
```diff
@@ -1316,6 +1317,102 @@ describe('morgan.compile(format)', function () {
...
+      return arg && url.parse(req.url)[arg]
+    }
+    var urlToken = morgan.url
+    morgan.token('url', token)
+    assert.equal(morgan.url, token)
+    morgan.token('url', urlToken)
+    done()
+  })
+
+  describe('should use the string `-` if return value of the token function is falsey', function () {
```
dougwilson: The `describe` should be keep below the `it` at the same level, that way they are ordered in the code in the way that mocha outputs it, which makes it so much easier to correlate them from the test output into the file. It's what we try to keep our style as.

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
...
+#### arguments to custom token formats
+
+Example of a custom token format that uses an argument. It will output the
+header related to the argument passed to the token.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+morgan.token('header', function getId (req, res, arg) {
```
dougwilson: I think the `getId` still being here is a typo.

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
...
+
+Sample app that will use a custom formatting function. This will print JSON
+instead of a simple string.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+var formatter = morgan.compile('[:date[clf]] :method :url :res[content-length]')
+var JSONOutput = function (morgan, req, res) {
```
dougwilson: The example should document as we want to describe: the first argument is `tokens`, not `morgan`.

dougwilson: Especially an issue since it's confusing when you're shadowing a variable (there is already a `var morgan = ` above, making it hard for users to sometimes understand what is actually being used here--the global or the argument).

dougwilson:
I'm not sure what is happening and where some of your comments are disappearing to; there is so much happening and every email for your comments simply link to `https://github.com/expressjs/morgan/pull/124` which is not helpful. Please forgive me if I'm not responding to everything, because some of your comments I just cannot find to respond to.

bmeck:
Gonna just take a step back and probably bail on this, seems a lot of
pushback.

dougwilson:
Sorry, reading back my comment I see what you mean. The "not helpful" was a comment to GitHub, in that the emails they are sending me are not helping me locate where your comment is in order for me to actually reply; it was not directed at you, I'm sorry if it came off that way :(

test/morgan.js
dougwilson: There is no public API to get tokens except the `tokens` argument from within a formatter. We can add an API to do this, if we think there is a use-case around having that.

test/morgan.js
dougwilson: Hm. I've been thinking on the "unclear why I need different ones" and I feel like I described it above, but obviously we have a disconnect here.. not sure how to proceed. Thoughts?

test/morgan.js
dougwilson: Cool. I don't doubt you changed the name, but I'm also not seeing a test called "a custom token function provided to .token(name,fn)"? What line is that on?

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
 }
 ```
 
+#### arguments to custom token formats
+
+Example of a custom token format that uses an argument. It will output the
```
indexzero: This is a solid point. I've had similar experiences on other projects. How about a more absurd example such as _"is it christmas?"_

> Example of a custom token format that uses an argument. It will output if the time
> of the current request is christmas.

```js

morgan.token('is-holiday', function getId (req, res, arg) {
  var date = new Date()
  var shortDate = `${date.getMonth() + 1}/${date.getDate()}`;
  var holiday = arg.toLowerCase().replace(/'/g, '');

  var holidays = {
    '1/1': 'new years day',
    '2/14': 'valentines day',
    '5/5': 'cinco de mayo',
    '10/31': 'halloween',
    '12/25': 'christmas',
    '12/31': 'new years eve'
  };

  var isHoliday = 'Nope';
  if (holidays[shortDate] === holiday) {
    isHoliday = 'Yup';
  }

  return String(isHoliday);
})

var app = express()

app.use(morgan('Is it Christmas? :is-holiday[christmas] :response-time'))

//
// Alternatively we could ask if it is Halloween
//
app.use(morgan('Is it Halloween? :is-holiday[halloween] :response-time'))


app.get('/', function (req, res) {
  res.send('hello, world!');
})
```

README.md
```diff
@@ -310,6 +325,54 @@ function assignId (req, res, next) {
...
+header related to the argument passed to the token.
+
+```js
+var express = require('express')
+var morgan = require('morgan')
+
+morgan.token('header', function getId (req, res, arg) {
+  return String(req.headers[arg])
```
indexzero: This is moot if we change the example to _"Is it Christmas?"_

README.md
indexzero: I think you're right @dougwilson. This sample should fail if I understand the shadowing correctly.

README.md
indexzero: The location makes sense to me. It is documenting the capabilities of the functions returned by `morgan.compile`

dougwilson: Ah. So maybe I'm just confused here. I was under impression from the issue that spawned this PR that the goal here was to document what the `format` argument in `morgan(format, options)` accepts. Are we missing that documentation and this is the documentation for `morgan.compile`? Since I mistook this as the documentation for the `format` argument, my thought here was that documentation for the `format` argument should be up in the section regarding the `format` argument, rather than in the `morgan.compile` section.

README.md
dougwilson: I mean, if we are struggling to even come up with a real example at all, then why even make an example?

dougwilson: How do you use this feature in your production systems? Just take that use-case (which is clearly real-world) and place it as the example?

README.md
indexzero: @dougwilson it is about documenting the `format` argument in `morgan(format, options)`. [On line 25 of README.md in this PR](https://github.com/expressjs/morgan/pull/124/files#diff-04c6e90faac2675aa89e2176d2eec7d8R25) @bmeck added the phrase:

> For the syntax of strings and functions see [morgan.compile](#morgancompileformat).

He then added this documentation regarding the semantics of functions returned by `morgan.compile` which is what is linked to by the documentation on line 25.

README.md
indexzero: @dougwilson how about this example which outputs a safe subset of process information? 

> Example of a custom token format that uses an argument. It will output the
property from `process` related to the argument passed to the token. The valid
arguments are restricted to safe properties of the `process` object in Node.js.

```js

morgan.token('process', function getId (req, res, arg) {
  var accessors = {
    arch: function () { return process.arch },
    pid: function () { return process.pid },
    platform: function () { return process.platform },
    title: function () { return process.title },
    uptime: function () { return process.uptime() },
    version: function () { return process.version }
  }

  if (accessors[arg]) {
    return String(accessors[arg]());
  }

  return String('Unknown arg: ' + arg);
})

var app = express()

app.use(morgan(':process[pid] :process[uptime] :response-time'))

app.get('/', function (req, res) {
  res.send('hello, world!')
})
```

README.md
dougwilson: Hm. I don't understand. I think I may just need to step back and process it. I see that change, but when I read through https://github.com/bmeck/morgan/blob/12230b264a41b04255e0d567ace2110276b5e62d/README.md just as a whole it didn't seem clear.

My process was I read the following:

> The format argument may be a string of a predefined name (see below for the names), a string of a format string, or a function that will produce a log entry.

Ok, predefined named, I'll see below for what that means. Oh, below it says "For the syntax of strings and functions see morgan.compile."

Ok, click that link.

> Compile a format string into a function for use by morgan

Cool.

> A format string is a string that represents a single log line and can utilize token syntax. Tokens are references by :token-name. If tokens accept arguments, they can be passed using [], for example: :token-name[pretty] would pass the string 'pretty' as an argument to the token token-name.

Cool. Good to understand about the tokens. I wonder what tokens are? I guess we'll get to that.

> Normally formats are defined using morgan.format(name, format), but for certain advanced uses, this compile function is directly available.

Ok, cool. Seems unrelated to the `format` argument I came to read about, but I can skip this line.

> The function returned takes arguments tokens , req, and res where tokens refers to morgan itself. If a log should be skipped the function will return null. The function returned or a custom function can be passed directly to morgan using morgan(myFn).

Cool, so that's what `morgan.compile` returns. But wait, the section is over. What do I give as the `format` argument? I know that `morgan.compile` takes a string as it's `format` argument, which is some token syntax thing, and that it returns a function that does some stuff.

Oh wait, I guess "The function returned or a custom function can be passed directly to morgan using morgan(myFn)." was just snuck there at the end, and I guess I can make the connection now.

Basically, my critism is that it's too 'round about a way. I'm sorry, I just don't see how this is useful documentation for the goal of just trying to get the `format` argument to the constructor across. It's just too much of a round about way.

If you strongly disagree, we can always wait for other Express.js members to chime in as well to see what we think collectively, but it just doesn't read well enough for me to want to merge in, personally.

indexzero:
@dougwilson took another stab at this on https://github.com/expressjs/morgan/pull/125

README.md
indexzero: Added this to #125

bmeck:
@dougwilson split into #125 and #127

dougwilson:
Thanks you two; I have taken a look at both PRs now. It seems like some of the review comments from this review got lost in the shuffle to new reviews, so if you re-read through here and see any that apply to your PR, that would help a lot! Otherwise I'll spend some time over this week cherry picking theones that still apply :)

test/morgan.js
bmeck: I'm opposed to this as changing one of those tests and not changing other could introduce issues.

test/morgan.js
bmeck: sounds fine, but I am just testing the actual existing behavior, if we sent the same argument length every time I would prefer that change to removing this test

PR #132, opened 2016-12-30
Add token to show date/time in current local timezone

miguelzf:
May be useful for applications targeting a particular geographical zone.


dougwilson:
Hi @miguelzf this has been discussed many times, even with multiple members of Express. You should simply define the token in your code directly. You can simply use Moment.js to make implementing the token a single line (and you can even re-define the `:date` token directly).

PR #133, opened 2017-01-24
Added keywords in package.json

badrmodoukh:
Hi, I added keywords to the package.json file so that it can get picked up by npmjs. These are the keywords I could think of for the morgan package. If you have any other keywords suggestions please let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,14 @@
     "Douglas Christopher Wilson <doug@somethingdoug.com>",
     "Jonathan Ong <me@jongleberry.com> (http://jongleberry.com)"
   ],
+  "keywords": [
+    "morgan",
+    "HTTP",
+    "logger",
+    "request logger",
+    "middleware",
+    "node.js"
+  ],
   "license": "MIT",
   "repository": "expressjs/morgan",
   "dependencies": {


dougwilson:
Thanks, @badrmodoukh , but since npm completely redid it's search engine, do the keywords even matter any longer? It looks like this module is already at the top for those keywords, so I'm not sure adding them will make a difference besides something else we have to maintain. Can you provide more information on exactly what uses them and how?

badrmodoukh:
Hi, @dougwilson, I am actually taking an introduction to open source course at college and one of our labs requires us to find a module on npmjs thats missing the keywords section. The purpose of this lab is just to get us introduced to the open source community since it is my first time contributing to the open source community. I am still new to node.js and open source so I'm not sure if my answer is right or wrong. 

So far what we have learned about the keywords section is that it's an optional field that you can include in the package.json file. Once these keywords are added it does not have to be maintained, it simply just gets displays in the keywords section on the npmjs website instead of displaying "None". The metadata in package.json file is used by the command-line tool npm as well as by the online registry npmjs. Providing good data for these tools makes it easier for developers to find and use node modules. I know that the morgan package is one of the top downloaded module and it would not matter if it doesn't contain the keywords section. Once again this lab is just to get us to participate in the open source community and get a feel to contributing to an open source project. 

Here is the link that the professor gave out explaining the keywords section: https://docs.npmjs.com/files/package.json#keywords

Also here is a screenshot of the keywords section for the morgan module on the npmjs online registry website
screen shot 2017-01-24 at 12 02 24 pm

PR #140, opened 2017-04-05
Add ability to silent morgan, useful when doing tests

rocketspacer:
I'm running mock tests with a dummy server, and find that too many logs are polluting my test results screen. Always wanted to have some mechanism to suppress log output.

Usage:
`morgan.silent = true`

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,7 @@ module.exports = morgan
 module.exports.compile = compile
 module.exports.format = format
 module.exports.token = token
+module.exports.silent = false
 
 /**
  * Module dependencies.
@@ -101,6 +102,10 @@ function morgan (format, options) {
   }
 
   return function logger (req, res, next) {
+
+    if (module.exports.silent)
+      return next()
+
     // request data
     req._startAt = undefined
     req._startTime = undefined
--- a/test/morgan.js
+++ b/test/morgan.js
@@ -1290,6 +1290,26 @@ describe('morgan()', function () {
       .expect(200, done)
     })
   })
+
+  describe('on silent mode', function () {
+
+    before(function () { 
+      morgan.silent = true 
+    })
+    after(function () {
+      morgan.silent = false
+    })
+    
+    it ('should not log any output', function (done) {
+      var stream = createLineStream(function () {
+        throw new Error('should not log line')
+      })
+
+      request(createServer({ format: 'default', stream: stream }))
+      .get('/')
+      .expect(200, done)
+    })
+  })
 })
 
 describe('morgan.compile(format)', function () {


rocketspacer:
@dougwilson 🍰

billouboq:
I don't really see an improvement here  : 

Since it's a setter, you still have to know when it's a test or a dev request, so it's not changing the problem.

If you want to silent morgan, just do : 

```js
if (process.env.NODE_ENV !== 'production' /* or whatever env variable you want to use */) {
   // or whatever thing you want to do
    app.use(morgan('dev'));
}
```

And launch your test by setting NODE_ENV to 'production' in command line.

rocketspacer:
@billouboq I'm writing tests and I don't  want them to affect current production code. Tests that require production code to change are bad tests. Can't leave too many artifacts around just because /* FOR THE PURPOSE OF TESTING */

billouboq:
@rocketspacer My bad was thinking about it locally and not globally.

So if you go server code : 

```js
var app = express()

app.use(morgan('dev'))
```

And in your test (which is in an other file) : 


```js
function beforeTest(next) {
  morgan.silent = true;
}

// do your test requests
```

this will silent morgan output?

rocketspacer:
@billouboq Yes, it will. I've also written a unit test for that.

The only down-side is that it will silent all middleware instances created by morgan, but making the silent option finer-grained (per middleware instance) can change the way we use morgan too much.

rocketspacer:
@billouboq any status on this ?

billouboq:
@rocketspacer Sorry I'm not in charge of this repo, I think @dougwilson is.

dougwilson:
Sorry, I lost track of this. I agree with @billouboq and I don't think it's the right solution. If you want to add it as an option to the constructor options, that would be better, but adding globals is bad, especially since the plan is to remove all the globals in the next major.

In addition, it is certainly possible you may need to disable other middleware for testing, so perhaps just make a generic disabling-middleware for your needs and you don't need to get every module on npm to incorporate any changes at all.

Otherwise, generally when you're having a problem when testing, it really is just weeding out bad practices in the programming itself. Ideally you should be using some form of dependency injection in your server to manage dependencies, including things like the logger or where the logger writes to.

For example, a properly constructed app server can easily test in logging silence, with the additional bonus that if it ever did need to add a test to validate that _logging actually happened_, that would also be possible:

`lib/app.js`
```js
const createLogger = require('./logger')
const express = require('express')

module.exports = function createApp (options) {
  const app = express()
  const logger = createLogger(options.logger || {})

  app.use(logger)
  // ... your stuff

  return app
}
```

`logger.js`
```js
const morgan = require('morgan')

module.exports = function createLogger(options) {
  const format = options.format || 'default'
  const immediate = options.immediate
  const stream = options.stream
  return morgan(format, { immediate, stream })
}
```

Now in your tests just construct your app with a null stream, for example:
```js
const createApp = require('../lib/app')

// ... later in testing
const app = createApp({ logger: { stream: new NullStream() } })
// ... use app in tests
```

This has many benefits, from using dependency injection to the tests not even needing to understand that `morgan` is what is creating the logs (so it can be swapped out if necessary) to issues you'll end up with if `require('morgan')` cannot even be resolved (or resolves to a different version) in the tests vs app code.

rocketspacer:
Ok, by wrapping morgan inside a logger adapter, I am now able to silent morgan by stubbing using sinon or just replace it with a next() middlewares when I run tests.

```
app.use(logger.logRequest);

// TESTS
require('logger').logRequest = (req, res, next) => next();
```

PR #128, opened 2016-10-20
Add objectMode option for objectMode streams

indexzero:
This is necessary if you want to log the results of `morgan` to another system which performs serialization itself. e.g. interop with other logging libraries like `winston` using `meta` instead of writing a single log line:

``` js
var winston = require('winston')
var morgan = require('morgan')
var express = require('express')

var app = express()

app.use(morgan(function (tokens, req, res) {
  return {
    method: tokens.method(req, res),
    url: tokens.url(req, res),
    status: tokens.status(req, res)
  }
}, {
  stream: function (meta) { winston.info('Request served', meta) },
  objectMode: true
})
```

without `objectMode: true` in the above example `winston` would output:

```
Request served [object Object]\n
```

Given that `objectMode` streams are a very common use-case for streams this seems like an essential addition to `morgan`.

PR #152, opened 2017-10-02
fixed additon of '\n' to non strings

oriben2:
I'm using morgan to write objects into a stream (winston). Because of the + '\n', morgan always writes '[object Object] \n'

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,7 +127,7 @@ function morgan (format, options) {
       }
 
       debug('log request')
-      stream.write(line + '\n')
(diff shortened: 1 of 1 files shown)


oriben2:
Ok, just noticed there's another pull request here on the same issue. Hope you merge one of these in the near future.


Thanks for the good work!

dougwilson:
Yea, I'm going to close as duplicate of #128 as this is basically trying to get at needing an object mode of operation.

PR #153, opened 2017-10-04
Added checks for unidentified format

SatioO:
I have added checks for format which is not specified in morgan. For eg. if user writes
`app.use(morgan("random"))` 
it is just printing random instead of logs in the console without any warnings.

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,6 +72,10 @@ function morgan (format, options) {
     deprecate('undefined format: specify a format')
   }
 
+  if (morgan[fmt] === undefined) {
+    deprecate('unidentified format: specify a proper format')
+  }
+
   // output on request instead of response
   var immediate = opts.immediate
 
--- a/index.js
+++ b/index.js
@@ -72,8 +72,11 @@ function morgan (format, options) {
     deprecate('undefined format: specify a format')
   }
 
-  if (morgan[fmt] === undefined) {
-    deprecate('unidentified format: specify a proper format')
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -72,6 +72,13 @@ function morgan (format, options) {
     deprecate('undefined format: specify a format')
   }
 
+  if (typeof fmt === 'string' && morgan[fmt] === undefined) {
+    if (fmt.split(':').length <= 1) {
+      deprecate('unidentified format: specify a proper format')
```
dougwilson: So if this is just a deprecation, it shouldn't change the behavior. It should just print that the provide value is now deprecated, but should still work the same, otherwise it's semver major and will wait to wait to land.

test/morgan.js
```diff
@@ -27,6 +27,22 @@ describe('morgan()', function () {
       .expect(200, cb)
     })
 
+    it('should throw warnings for unspecified format', function (done) {
```
dougwilson: I'll take a look into this test to see how I can help, but this test doesn't actually do anything that is being asserted: for example it's not asserting anywhere what the format was changed to `'default'` and it doesn't assert that any warning was actually displayed.

dougwilson:
I'm going to close this as stale as I cannot actually push up a fix to this PR directly, as I don't seem to have the permissions. If you would like to address the above unresolved review comments, please feel free to do so and just open the PR a-new :)

PR #156, opened 2017-11-18
Make Morgan easier to use with web frameworks that do not have middleware

TimeBomb:
I recently successfully and pretty painlessly implemented morgan into micro - https://github.com/zeit/micro.
The implementation looks something like this:

```
function(req, res) { // micro's http server function
    morgan('short')(req, res, function() {});
}
```

See the empty function? Morgan throws an exception if I don't pass it a `next` function,  which does not exist in micro.

This PR has morgan only calling `next()` if it is passed to morgan, which should remove the necessary empty function for web frameworks without middleware support, while not harming any preexisting implementations. Let me know what you think.

### DIFF
--- a/index.js
+++ b/index.js
@@ -141,7 +141,9 @@ function morgan (format, options) {
       onFinished(res, logRequest)
     }
 
-    next()
+    if (next) {
+      next()
+    }
   }
 }
 


dougwilson:
Sounds like there should just be a module that would wrap standard middkeware like this one to alter it to the interface of the micro framework, otherwise you are essentially asking for all middleware out there to change how they work. Why not take that generic approach?

TimeBomb:
Thanks for the different perspective.

>   otherwise you are essentially asking for all middleware out there to change how they work
Definitely not a good way of going about things.

My initial reasoning behind this was from the perspective "This is a really simple improvement that won't harm anything, only help", but as you put it so succinctly, this is not the best angle to take on the issue. Something along the lines of an adapter for middleware -> micro makes much more sense.

Cheers.

dougwilson:
Yea, I'm certainly not saying there is an actual issue with this change here at all, but especially if you have favorite middleware different authors may think otherwise, so the other benefit is that you can use the middleware without each middleware author buying into it as well.

PR #160, opened 2018-02-21
use x-forwarded-for so we get the real ip when the server is behind a…

mihaisecasiu:
… proxy

### DIFF
--- a/index.js
+++ b/index.js
@@ -463,7 +463,7 @@ function getFormatFunction (name) {
  */
 
 function getip (req) {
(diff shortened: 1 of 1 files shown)


dougwilson:
You're expected to setup `req.ip` or redefine the token to use your own logic: https://github.com/expressjs/morgan#remote-addr

ledlamp:
just add `app.enable('trust proxy');` and express will set `req.ip` to the real ip from `X-Forwarded-For` headers.

PR #164, opened 2018-05-06
[Fixes #163] Don't propose "default" format when provided with an Object as fmt

### LINKED ISSUE #163: Inconsistent defaults and deprecation messages
Hi guys !

This is probably a really tiny not-too-imporant bug, but well.. just found it and I think that it could help others.

Setting up express with morgan in this way

```js
const app = express()
app.use(morgan)
```

Gives this deprecation warning (when trying to log a request, not at start up)

```bash
morgan deprecated morgan(options): use morgan("default", options) instead node_modules/express/lib/router/layer.js:95:5
morgan deprecated default format: use combined format node_modules/express/lib/router/layer.js:95:5
```

It is of course wrong, but I believe it was "the old way". 
Anyway notice there that id proposes you to use `default`  (`use morgan("default", options)`)

But this is in turn deprecated, so if you change it like that you will get another deprecation message

```js
const app = express()
app.use(morgan('default'))
```
You get

```bash
morgan deprecated default format: use combined format index.js:30:30
```

So, I believe that the first message should be updated to directly instruct you to use the "combined" option instead of the "default".

Thanks !
Regards !

javierfernandes:
Changing default usage of the "default" format, which is deprecated to suggest using the "combined" format.

This fixes this case `morgan({})`
It was proposing you to use `morgan('default')` which is deprecated. Now it says

```
morgan deprecated morgan(options): use morgan(format, options) instead
```

# Testing

I've added a new kind of test (`/test/deprecations.js`) because `/test/morgan.js` was suppressing all deprecation message. 
There was no other test trying to capture and assert deprecations.

Also, I believe that this new kind of test could help this other PR https://github.com/expressjs/morgan/pull/153 that was criticized because it was not asserting about the deprecation message.

# Note

Notice also that this doesn't actually handle the problem of trying to use morgan in a wrong way, using it directly as a middleware instead of calling it as a function to get the middleware.
That can be done (with one of this new kind of test :P) introducing a new check here

```
  if (format && typeof format === 'object') {
    opts = format
    fmt = opts.format || 'combined'

    // smart deprecation message
    if (format.baseUrl) {
       deprecate('morgan(options): instead of "app.use(morgan)", you must use morgan like this "app.use(morgan(\'combined\')"')
    } else {
      deprecate('morgan(options): use morgan(' + (typeof fmt === 'string' ? JSON.stringify(fmt) : 'format') + ', options) instead')
  }
```
Note that `format.baseUrl` is basically just a way to try to recognize that the first arg is express "request" object, which means that you have used morgan fn directly as a middleware. Maybe there is a better way to check that :)

I'll leave that to you or another PR, since I'm not sure about the deprecation message.
And actually I'm not sure that case should be handle as a deprecation warning message. I think that in that case the whole thing should blow up with an exception, because it won't probably work at all. Right ??

Cheers !!

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,7 +62,7 @@ function morgan (format, options) {
 
   if (format && typeof format === 'object') {
     opts = format
-    fmt = opts.format || 'default'
+    fmt = opts.format || 'combined'
 
     // smart deprecation message
     deprecate('morgan(options): use morgan(' + (typeof fmt === 'string' ? JSON.stringify(fmt) : 'format') + ', options) instead')
@@ -446,7 +446,7 @@ function format (name, fmt) {
 
 function getFormatFunction (name) {
   // lookup format
-  var fmt = morgan[name] || name || morgan.default
+  var fmt = morgan[name] || name || morgan.combined
 
   // return compiled format
   return typeof fmt !== 'function'
--- a/index.js
+++ b/index.js
@@ -62,10 +62,11 @@ function morgan (format, options) {
 
   if (format && typeof format === 'object') {
     opts = format
-    fmt = opts.format || 'combined'
+
+    fmt = opts.format || 'default'
 
     // smart deprecation message
-    deprecate('morgan(options): use morgan(' + (typeof fmt === 'string' ? JSON.stringify(fmt) : 'format') + ', options) instead')
+    deprecate('morgan(options): use morgan(' + (typeof opts.format === 'string' ? JSON.stringify(opts.format) : 'format') + ', options) instead')
   }
 
   if (fmt === undefined) {
@@ -446,7 +447,7 @@ function format (name, fmt) {
 
 function getFormatFunction (name) {
   // lookup format
-  var fmt = morgan[name] || name || morgan.combined
+  var fmt = morgan[name] || name || morgan.default
 
   // return compiled format
   return typeof fmt !== 'function'
--- a/test/deprecations.js
+++ b/test/deprecations.js
@@ -33,7 +33,7 @@ describe('morgan() deprecations', function () {
   })
 })
 
-var ANSI_REGEX = new RegExp(/\u001b\[.*?m/, 'g')
+var ANSI_REGEX = /\u001b\[.*?m/g
 function removeANSI (string) {
   return string.replace(ANSI_REGEX, '')
 }
--- a/test/deprecations.js
+++ b/test/deprecations.js
@@ -20,11 +20,18 @@ describe('morgan() deprecations', function () {
       process.stderr.write = originalConsoleError
     }
   })
+  afterEach(function () {
+    errors = []
+  })
 
   it('should tell you to use morgan "combined" if format arg is an object', function () {
     morgan({})
-    assert.equal(getErrors().length, 1)
-    assert(startsWith(getErrors()[0], 'morgan deprecated morgan(options): use morgan("combined", options) instead'))
+    assert(startsWith(getErrors()[0], 'morgan deprecated morgan(options): use morgan(format, options) instead'))
+  })
+
+  it('should tell you to use "combined" instead of "default"', function () {
+    morgan('default')
+    assert(startsWith(getErrors()[0], 'morgan deprecated default format: use combined format'))
   })
 
   it('regex removes all ANSI characters', function () {
--- a/test/deprecations.js
+++ b/test/deprecations.js
@@ -5,7 +5,7 @@ describe('morgan() deprecations', function () {
   var originalConsoleError
   var errors = []
   function getErrors () {
-    return errors.map(e => removeANSI(e.toString('utf8')))
+    return errors.map(function (e) { return removeANSI(e.toString('utf8')) })
   }
   before(function () {
     originalConsoleError = process.stderr.write
--- a/test/deprecations.js
+++ b/test/deprecations.js
@@ -11,7 +11,7 @@ describe('morgan() deprecations', function () {
     originalConsoleError = process.stderr.write
 
     process.stderr.write = function write (chunk, encoding) {
-      errors.push(Buffer.from(chunk, encoding))
+      errors.push(Buffer.from ? Buffer.from(chunk, encoding) : new Buffer(chunk, encoding))
     }
   })
   after(function () {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -446,7 +446,7 @@ function format (name, fmt) {
 
 function getFormatFunction (name) {
   // lookup format
-  var fmt = morgan[name] || name || morgan.default
+  var fmt = morgan[name] || name || morgan.combined
```
javierfernandes: I changed this one also thinking that if "default" is considered as deprecated then it shouldn't be used as a default, right ?

But, although I could create a unitary test for this function along, it didn't make much sense. Then I was not able to create a particular test for this change. I just make sure all other tests kept running fine.

But it would be great to get a confirmation about this change.
Thanks !

dougwilson: Yea, this change will break people, so you'll have to determine a different way to make the change. The formats are slightly different, which is why one is deprecated and didn't just change the default without a message if that makes sense.

index.js
```diff
@@ -62,7 +62,7 @@ function morgan (format, options) {
 
   if (format && typeof format === 'object') {
     opts = format
-    fmt = opts.format || 'default'
+    fmt = opts.format || 'combined'
```
dougwilson: So this will break existing logs due to the format change, right? The date formats are different between default and combined, so you can't just change the default value of format.

dougwilson:
> And actually I'm not sure that case should be handle as a deprecation warning message. I think that in that case the whole thing should blow up with an exception, because it won't probably work at all. Right ??

Yes, though this is a fundamental issue you can do to any middleware module out there. Perhaps there is a change you can make to Express on this to catch the mistake instead of needing to change every single middleware that exists with whatever the check they should use is. That would be the biggest win for everyone :+1:

test/deprecations.js
```diff
@@ -0,0 +1,50 @@
...
+  var originalConsoleError
+  var errors = []
+  function getErrors () {
+    return errors.map(function (e) { return removeANSI(e.toString('utf8')) })
+  }
+  before(function () {
+    originalConsoleError = process.stderr.write
+
+    process.stderr.write = function write (chunk, encoding) {
+      /* eslint node/no-deprecated-api: 0 */ // this is actually trying to be compat with older node
```
dougwilson: Rather than disabling it, I would follow the suggestion of the linter and use `safe-buffer` instead.

javierfernandes: Ok, I thought you wouldn't like to add a new package just for this.

test/deprecations.js
```diff
@@ -0,0 +1,50 @@
...
+var assert = require('assert')
+
+describe('morgan() deprecations', function () {
+  var originalConsoleError
+  var errors = []
+  function getErrors () {
+    return errors.map(function (e) { return removeANSI(e.toString('utf8')) })
+  }
+  before(function () {
+    originalConsoleError = process.stderr.write
```
dougwilson: Personally, I don't bother writing tests for these deprecation messages at all, but if you feel so inclined, I would say please don't do it by replacing `process.stderr.write` -- mocha itself may try to write there. Instead capture the messages through a listener: https://www.npmjs.com/package/depd#processondeprecation-fn

test/deprecations.js
dougwilson: It's just a devDependency since this is a test.

test/deprecations.js
```diff
@@ -0,0 +1,50 @@
...
+  after(function () {
+    if (originalConsoleError) {
+      process.stderr.write = originalConsoleError
+    }
+  })
+  afterEach(function () {
+    errors = []
+  })
+
+  it('should tell you to use morgan "combined" if format arg is an object', function () {
```
dougwilson: I think there is a typo somewhere as the description doesn't match the actual test. From looking at it, I think the description is what just needs the update 👍

index.js
```diff
@@ -62,10 +62,11 @@ function morgan (format, options) {
 
   if (format && typeof format === 'object') {
     opts = format
+
     fmt = opts.format || 'default'
 
     // smart deprecation message
-    deprecate('morgan(options): use morgan(' + (typeof fmt === 'string' ? JSON.stringify(fmt) : 'format') + ', options) instead')
+    deprecate('morgan(options): use morgan(' + (typeof opts.format === 'string' ? JSON.stringify(opts.format) : 'format') + ', options) instead')
```
dougwilson: Nice! This makes a lot of sense -- it's basically just telling you to use `(format, options)` instead of echoing a format you didn't provide. That makes a lot of sense.

test/deprecations.js
javierfernandes: I took with from https://github.com/dougwilson/nodejs-depd/blob/master/test/support/capture-stderr.js

Maybe you should also fix that :)

dougwilson: No, because that is the module itself, to validate its writing to stderr. Modules embedding depd should use the listener hook only (i.e. they should follow the README).

javierfernandes:
(couldn't reply to the comment inlined)

```
Yes, though this is a fundamental issue you can do to any middleware module out there. Perhaps there is a change you can make to Express on this to catch the mistake instead of needing to change every single middleware that exists with whatever the check they should use is. That would be the biggest win for everyone
```

That is not possible unless you are in a typed language. Express `app.use()` expects a function, and morgan is a function as well as the return value from morgan().
There is no way express can check that. Not even checking cardinality because I that would restrict you from doing something like `app.use( (...args) => )`

dougwilson:
Then if it's not possible to fix, then I think it's best left to the README to show the proper usage. Like I said this issue can happen to any middleware, and if the solution is to go and pull request every single middleware out there, then I am a big :-1: on that and won't take a change in this module for that. If the solution is to used a typed langauge, then you can use typescript which would have caught this.

test/deprecations.js
javierfernandes: Ok, Please notice that your personal decision of not having tests for deprecations is what actually made it possible to have the issue that I reported initially. If you don't test it then errors on those message will get through.

You are too picky dude. It is ok, nothing personal, but I don't have more time to spend on this, that's why I closed it. I just wanted to help but I think that we are both losing time here. If you do it yourself then that would be faster.

Good luck there !

PR #165, opened 2018-05-19
Record response time at response end instead of response start

bobwei:
Should record response time at response end instead of response start so that we can record whole response time for streaming response.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 unreleased
 ==========
 
+  * Add `:total-time` token
   * Fix trailing space in colored status code for `dev` format
   * deps: basic-auth@~2.0.1
      - deps: safe-buffer@5.1.2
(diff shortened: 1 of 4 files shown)


dougwilson:
Also the docs need to be updated (though since this should be a new option,  add new docs). Here is the current docs:

> :response-time[digits]
> The time between the request coming into morgan and when the response headers are written, in milliseconds.

dougwilson:
Hi @bobwei sorry I never came back to this PR. I just realized that instead of landing this as a flag that changes the behavior of an existing token, we should modify the PR to just add a new token with the behavior, which same the same ultimate outcome, but the benefit of being able to include both of these times in the output if that's desired.

dougwilson:
I have just written all the code to move this to a `:total-time` token. If you are agreeable on this approach, I can squash this into your PR here and merge as your PR still.

bobwei:
Hi @dougwilson, sure, please proceed.

pjsg:
Can this be merged please?

bobwei:
Go go

PR #174, opened 2018-07-14
Add colored status token

### LINKED ISSUE #171: Add :status field with color as a default
I love the `dev` config, only I want it to have a timestamp as well. 

I know that if i copy the dev config manually, I can simply add `:date[iso]`, but then I lose the awesome colors of the response codes.

Can you either add another default format which is the `dev` config with a timestamp or can you simply make a `:status` field that by default has color? e.g. `:colored-status`

st-sloth:
Coloring status makes sense for whatever format user wants, not just the internal `dev` one. So using `:status-colored` gives that ability.

Had to edit tests because, in master, status coloring is being disabled after a space following status specifically in dev format. Now coloring is disabled right after status text.

As for collaterally removed memoization for `dev` format, well, wouln't it be better to memoize all formats that go through `compile`, not just the `dev` one? If that's the case, I'll do that separately.

resolves #171


dougwilson:
Hi @st-sloth sorry I never saw your PR. I'm currently going through the repo and cleaning up the issues / PRs which is how I found it. So I think this is a good idea, though it just needs a rebase to fix the conflicts after I merged a couple other PRs.

As for your memoize question, the compile is a public function and since memoization has no expiration and currently users could be generating a new format on every request for who knows what reason, it is unlikely to be safe to suddenly add an unbounded cache to this function.

But maybe I'm completely misunderstanding your question.

tahv0:
Will this get merged soon?

Inkbug:
Any updates? I implemented this manually for my self, but it would be much more elegant if this could be in the original base.

ryhinchey:
@st-sloth are you still working on this? I picked it up in this PR and am working on updating the tests https://github.com/expressjs/morgan/pull/227

st-sloth:
@ryhinchey No, I'm done and have been waiting for a reply or merge ever since. Happy to see it picked up, revived and bettered, thank you! Good luck with your PR!

Vivida1:
Any updates on this? This would be really helpful.

adamreisnz:
Still not merged..?

PR #175, opened 2018-07-18
Color :status by default

toymachiner62:


toymachiner62:
There are currently some failing unit tests that i'm not exactly sure how to fix. Mostly on asserting colors. Could use some help on this.

PR #182, opened 2018-10-01
Simplify log file rotation example

davidjb:
The rfs module will automatically create the log directory (https://www.npmjs.com/package/rotating-file-stream#function-filenameindex) so this example doesn't need to manually check for or create the directory.

From https://www.npmjs.com/package/rotating-file-stream: “Note: if part of returned destination path does not exists, the rotation job will try to create it.”

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,21 +312,16 @@ file per day in the `log/` directory using the
 
 ```js
 var express = require('express')
-var fs = require('fs')
 var morgan = require('morgan')
 var path = require('path')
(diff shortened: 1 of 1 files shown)


dougwilson:
Awesome, thanks! Looks like just a lint issue in the example:

> 'fs' is assigned a value but never used

davidjb:
@dougwilson Updated and pushed accordingly, thanks for catching that.

PR #184, opened 2018-11-20
new line format

nllahat:
Morgan prints new line at the end of every log. It causes our AWS logs full of empty lines and makes it hard to detect and trace bugs. That's why I added an option - 'newLineFormat' . If this option does not exist - the fallback is '\n'. 
Our use case will be {newLineFormat: ''}


dougwilson:
This is by design.

PR #187, opened 2019-02-19
docs: Fix doc due to return undefined

koooge:
Hello. This PR just fix doc. morgan returns status = undefined, not '-'.
https://github.com/expressjs/morgan/blob/master/index.js#L266

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ The status code of the response.
 
 If the request/response cycle completes before a response was sent to the
 client (for example, the TCP socket closed prematurely by a client aborting
-the request), then the status will be empty (displayed as `"-"` in the log).
+the request), then the status will be empty.
 
 ##### :url
 


dougwilson:
All falsy values are displayed as `-` in the logs.

gms-ta-koge:
Hmm, in my env with express

```
  return [
    tokens.method(req, res),
    tokens.url(req, res),
    tokens.status(req, res),
    tokens.res(req, res, 'content-length'), '-',
    tokens['response-time'](req, res), 'ms'
  ].join(' ')
}));

app.use('/foo', (req, res) => {
  // nop
});
```

outputs
`GET /foo   -  ms`

tokens.status is not '-', right?

koooge:
Hmm, in my env with express

```
  return [
    tokens.method(req, res),
    tokens.url(req, res),
    tokens.status(req, res),
    tokens.res(req, res, 'content-length'), '-',
    tokens['response-time'](req, res), 'ms'
  ].join(' ')
}));

app.use('/foo', (req, res) => {
  // nop
});
```

outputs
`GET /foo - ms`

tokens.status is not '-', right?

dougwilson:
The note you removed is

> displayed as `"-"` in the log

That is not documenting the actual return value of the function.

dougwilson:
If you provide a custom function it will display however you make the string. Try this instead:
```
app.use(morgan(':method :url :status :res[content-length] :response-time ms'));

app.use('/foo', (req, res) => {
  // nop
});
```

koooge:
Thank you for your quick response. I will try it 🙇

PR #191, opened 2019-04-12
Refactoring

wesleimp:


dougwilson:
Thanks for your pull request, but I don't think it's necessary, especially since we are already happy with the current organization and this pull requests no longer passes our CI and reduces the code coverage.

Maybe if you can ellaborate on what specifically drove you to make this change and articulate what the benefits are compared to the current wording that would help us understand.

I haven't specifically looked though the new code yet as I wanted to better understand the above first so we can drive together in the right direction.

dougwilson:
Closing due to no response.

Issue #181, opened 2018-09-27
is there a way to pass a 'meta' object to winston through morgan?


rconstantine:
I'm using Morgan for the HTTP logging and Winston wraps it. There is no explicit link between the two other than setting up the 'stream' option. Winston allows for a 'meta' object that can contain additional fields beyond the message and level. All of my logs except the Morgan ones contain metadata, like the username of the person triggering the leg event (or SYSTEM if not a user), and I'd like to add the same kind of thing for Morgan logs going to Winston. Any ideas?

dougwilson:
So morgan is just a text logger and writes just a string to a stream for each log entry. I'm not sure what you're really talking about, so hard to answer more than that. Perhaps a code example would help clarify.

rconstantine:
OK, that means there's no option. That's OK. The data is still available, I just can't parse it easily. I'll have to consider formatting as JSON. Thanks for the reply.

Uko:
@rconstantine did you find a nice solution? Because I need to log objects (yes, like JS objects, not text), and I'm considering to hack Morgan to be able to specify a function that should be executed for each obtained object. If you know about other solutions it would be nice to learn :)

Uko:
Actually, it's really interesting. If there was no `+ '\n'` in https://github.com/expressjs/morgan/blob/master/index.js#L130 it could be straightforward to do object logging.

tonyxiao:
The simplest solution would be to change this one line

https://github.com/expressjs/morgan/blob/master/index.js#L130

from `stream.write(line + '\n')` to `stream.write(line + '\n', req, res)`. That way additional context related to the request can be passed to the stream object if needed. Would anyone object to a fork?

dougwilson:
Hi @tonyxiao you're certainly welcome to fork it. As far as your proposal goes, the `.write` function takes the values defined by the Node.js stream interface: https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_writable_write_chunk_encoding_callback

That is, the signature is `.write(chunk[, encoding][, callback])`. You would end up passing the `req` object as the `encoding` and the `res` object as the `callback` in your case.

PR #194, opened 2019-05-07
Added enableObjectStream option

SnakeA:
## Background 

I've got a similar use case with the issue raised: https://github.com/expressjs/morgan/issues/181. 

I am using Morgan for a lot of my projects and I would like to be able to write an Object in a stream instead of just a string.

Reason is that I am using Winston which supports meta-data for reach-er logs. I would like to use morgan and pass an object to Winston - instead of parsing a string. 

## Changes
- Added a **backwards compatible** `enableObjectStream` in morgan() call. 
 - Defaults to false
- It needs a custom format function and a defined stream to work.
- The only diff is that if the above are true (`customFormatFunction && stream && enableObjectStream`), we are not concatenating a `\n` (newline char), hence morgan can stream an object.

Any improvements/changes are welcome.

TODO:
- Update README
- Update HISTORY

but I'd like to get some opinions first 😄

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,12 +79,14 @@ function morgan (format, options) {
   var skip = opts.skip || false
 
   // format function
-  var formatLine = typeof fmt !== 'function'
+  var isFormatFunction = typeof fmt === 'function'
+  var formatLine = !isFormatFunction
     ? getFormatFunction(fmt)
     : fmt
 
   // stream
   var buffer = opts.buffer
+  var isStreamDefined = !!opts.stream
   var stream = opts.stream || process.stdout
 
   // buffering support
@@ -126,8 +128,13 @@ function morgan (format, options) {
         return
       }
 
-      debug('log request')
-      stream.write(line + '\n')
+      if (opts.enableObjectStream && isStreamDefined && isFormatFunction) {
+        debug('log request without newline character')
+        stream.write(line)
+      } else {
+        debug('log request')
+        stream.write(line + '\n')
+      }
     };
 
     if (immediate) {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "on-headers": "~1.0.1"
   },
   "devDependencies": {
-    "eslint": "5.9.0",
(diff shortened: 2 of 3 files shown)


SnakeA:
@dougwilson  Could you please have a look 😄  Thanks !

dougwilson:
I will after the express 4.17 release is out, as there was a premature vulnerability listed and it is our number 1 priority right now, I hope you understand.

SnakeA:
@dougwilson 
Thank you for getting back to me 😄 No worries I do understand that you have other priorities 

I did some digging and I've realised that a lot of people tried to make this change to Morgan but it never went out. 

Morgan is being used by numerous of our project and does the job, other than the fact that it emits a string that we have to parse. 

It would be great if we can work something out, otherwise, we might have to implement our own middleware 

Thanks again and looking forward to your review & opinion

SnakeA:
@dougwilson 

Any updates? 
I can see that express 4.17 was released last week? 

Thanks!

dougwilson (changes requested):
Sorry, still getting reports about 4.17 from users. Here is a very quick review for points to work on:

1. No docs for new feature in pr
2. Does not work right when buffer feature enabled
3. Silently does nothing when using built in formatters. It could throw an error or at least be documented. 

I hope this feedback helps while I am working on 4.17 issues in parallel.

UlisesGascon:
Thanks for opening this, @SnakeA, and sorry it sat for so long. Closing it now because the use case has been addressed on master by #375.

Rather than a new option, morgan now checks the stream's own `writableObjectMode` flag. When it is `true` and a custom format function returns an object, the object is written to the stream as is, with no newline appended. String formats keep producing strings, so nothing changes for existing users.

Example:

```js
var stream = new Writable({
  writableObjectMode: true,
  write: function (obj, encoding, callback) {
    logger.info('request', obj)
    callback()
  }
})

  return { method: tokens.method(req, res), url: tokens.url(req, res), status: tokens.status(req, res) }
}, { stream: stream }))
```

PR #195, opened 2019-05-13
comment deprecation warning

dhowe:


dougwilson:
This is a breaking change.

dougwilson:
Please see my comment in the main issue thread. This was deprecated almost 5 years ago without any issues. Perhaps if you can provide a pull request that will both silence the warning for your use-case and keep it intact for the folks who are using something that will be removed, that would be more productive to help resolve the issue.

As I said, I do intent to look into it, but I haven't had a chance yet.

PR #197, opened 2019-05-23
feat: add locale date string

xrr2016:
Need locale date string.


dougwilson (changes requested):
Thank you for your pull request! Sorry it fell through the cracks, as it were. Can you elaborate more on the use-case for this? Also please add docs and tests for this new feature if you would like to see it land.

UlisesGascon:
Hi @xrr2016! If you need help with the tests. Please ping me and I will try to help you 👍

PR #213, opened 2019-11-04
Fix deprecation warning for node12 LTS

zckrs:
https://nodejs.org/api/deprecations.html#deprecations_dep0066_outgoingmessage_prototype_headers_outgoingmessage_prototype_headernames

Each time this function is called, we have a warning in console output.

`(node:21443) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated`

### DIFF
--- a/index.js
+++ b/index.js
@@ -478,8 +478,12 @@ function getip (req) {
  */
 
 function headersSent (res) {
+  var isHeadersAlreadySent = res.getHeaders
+    ? res.getHeaders()
+    : res._headers
+
   return typeof res.headersSent !== 'boolean'
-    ? Boolean(res.getHeaders())
+    ? Boolean(isHeadersAlreadySent)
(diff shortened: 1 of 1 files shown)


dougwilson:
This would only occur if you are using an old version of morgan. Running the current version of morgan and Node.js 12 does not produce that for me. Without tests to confirm this fixes the issue included, can you outline the exact method to manually reproduce the issue so I can review the change?

dougwilson:
Please don't forget to provide the steps to reproduce the deprecation message so we can land this soon for you. I just tried the example in the README running under Node.js 12 and there was no deprecation message with he current version, so there must be something more to the issue than the description leads to.

zckrs:
I try to reproduce with small code then you are right this not related to morgan directly. Sorry for noise.

PR #219, opened 2020-01-02
Add record json response option

qianfeiqianlan:
Hello, I have added `recordJsonResponse` option to record the response body if this response with `application/json` header.
I think if I want to record a response in log, this is necessary.
The reference for this implementation: [https://github.com/bithavoc/express-winston](url)

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,10 @@ morgan('combined', {
 })
 ```
 
+##### recordJsonResponse
+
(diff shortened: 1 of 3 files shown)


dougwilson:
Thank you, your work is great, but this feature is outside the scope of the standard built-in tokens. You should be able to implement this in your app without any modification needed to morgan itself. If you're not sure how, please feel free to ask and we can assist :+1:

PR #221, opened 2020-02-21
Updating code example for logging on rotation

seven-deuce:
'rotating-file-stream' api has changed and most people will install the latest. So the example should adapt to that.
Also the previous code example was incomplete and left many people wondering about details of implementation.
My example is complete and short. Anyone can copy-paste it and get it immediately working. No need to go to the 'rotating-file-stream' library and read its documentation!

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,17 +320,17 @@ var app = express()
 
 // function for creating names for files
 const generateName = (date) => {
-return new Date(date)
-		.toLocaleDateString()
-		.replace(/\//g, "-")
-		.concat(".log")
+  return new Date(date)
+    .toLocaleDateString()
+    .replace(/\//g, '-')
+    .concat('.log')
 }
 
 // create a rotating write stream
 const accessLogStream = rfs.createStream(generateName, {
-	interval: "1d", // rotate daily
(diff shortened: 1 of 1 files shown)


dougwilson:
> Also the previous code example was incomplete and left many people wondering about details of implementation.

Can you elaborate on that? What details were missing? Just reading the changes here, I'm still not sure what exactly is going on in that regard.

> My example is complete and short. Anyone can copy-paste it and get it immediately working. No need to go to the 'rotating-file-stream' library and read its documentation!

I'm confused, as I believe the current documentation is already that way? At least, I copied out the example and it worked for me without modification. Can you elaborate on this point?

dougwilson:
Thank you for the feedback. That all sound good, but sounds like feedback to make to the `rotating-file-stream` about the default options. I would rather keep the example simple, and let users expand it as they need. The example is just to get it to work.

On your rotation note, replacing the same file is still rotating the file; it sounds like your need it to keep the older data, which is fine, but not everyone needs that.

The goal of the example is to use as little non-morgan code as possible to show _how to use an external stream with morgan_.

PR #226, opened 2020-02-27
mocking date for a test to ensure pad2 runs with single digit dates

ryhinchey:
This PR adds a test to always make sure `pad2` is covered.  When running tests, if every value passed to `pad2` is two digits (which depends on the time the tests are run), a drop in coverage can happen. 

To mock the date I added `sinon@4.0.0.` as a dependency.  `4.0.0` was used as it was the last version I could find that had an `engines` field for the versions of node this library has to support: https://github.com/sinonjs/sinon/blob/v4.0.0/package.json#L91-L93

### DIFF
--- a/test/morgan.js
+++ b/test/morgan.js
@@ -9,6 +9,7 @@ var join = require('path').join
 var morgan = require('..')
 var request = require('supertest')
 var split = require('split')
+var sinon = require('sinon')
 
 describe('morgan()', function () {
   describe('arguments', function () {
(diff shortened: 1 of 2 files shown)


test/morgan.js
```diff
@@ -211,6 +211,27 @@ describe('morgan()', function () {
           .expect(200, cb)
       })
 
+      it('should handle dates with single digits when using "clf" format', function (done) {
+        var GlobalDate = global.Date
+
+        global.Date = function () { return new GlobalDate('2020-02-27') }
```
ryhinchey: I'm not proud of this but `sinon` did not play nice with node 0.8 or 0.10 https://travis-ci.org/expressjs/morgan/builds/655979079

PR #227, opened 2020-02-27
Colored status

ryhinchey:
This PR builds upon https://github.com/expressjs/morgan/pull/174 and adds more tests to up the coverage

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ Same as `:status` but the output will be colored red for server error codes,
 yellow for client error codes, cyan for redirection codes, green for success codes,
 and uncolored for all other codes.
 
-This token also adds a reset escape code to ensure color will only be applied to the status code.  The escape code removes any existing colors in your logs that come after the `:status-colored` token. For something like a background color that's applied to the entire log string, you must add it back after the `:status-colored` token.
+In addition, this token adds a reset escape code to ensure that the color will only be applied to the status code.  Due to how ANSI escape codes work, any existing colors in your logs that come after the `:status-colored` token will be removed. 
 
 For example, this log format sets a magenta background color to the entire log string.  Notice that it's added at the beginning of the log and immediately after the `:status-colored` token.
 
--- a/README.md
+++ b/README.md
@@ -110,13 +110,13 @@ Standard Apache common log output.
 
 ##### dev
 
-Concise output colored by response status for development use. The `:status`
+Concise output colored by response status for development use. The `:status-colored`
 token will be colored green for success codes, red for server error codes,
 yellow for client error codes, cyan for redirection codes, and uncolored
 for information codes.
 
 ```
-:method :url :status :response-time ms - :res[content-length]
+:method :url :status-colored :response-time ms - :res[content-length]
 ```
 
 ##### short
@@ -219,6 +219,18 @@ has finished being written out to the connection, in milliseconds.
 
 The `digits` argument is a number that specifies the number of digits to
 include on the number, defaulting to `3`, which provides microsecond precision.
+
+##### :status-colored
+
+Same as `:status` but the output will be colored red for server error codes,
+yellow for client error codes, cyan for redirection codes, green for success codes,
+and uncolored for all other codes.
+
+This token also adds a reset escape code to ensure color will only be applied to the status code.  The escape code removes any existing colors in your logs that come after the `:status-colored` token. For something like a background color that's applied to the entire log string, you must add it back after the `:status-colored` token.
+
+For example, this log format sets a magenta background color to the entire log string.  Notice that it's added at the beginning of the log and immediately after the `:status-colored` token.
+
+`morgan('\x1b[45m:method :url :status-colored\x1b[45m :res[content-length]\x1b[0m')`
 
 ##### :url
 
--- a/index.js
+++ b/index.js
@@ -48,13 +48,6 @@ var CLF_MONTH = [
 var DEFAULT_BUFFER_DURATION = 1000
 
 /**
- * Placeholder that replaces empty token values in resulting log line
- * @private
- */
-
-var EMPTY_TOKEN_VALUE_PLACEHOLDER = '-'
-
-/**
  * Create a logger middleware.
  *
  * @public
@@ -280,10 +273,7 @@ morgan.token('status-colored', function getStatusColoredToken (req, res) {
         : status >= 200 ? 32 // green
           : 0 // no color
 
-  // using placeholder here since `compile` adds it if token function returns falsy,
-  // but this coloring token returns color data anyway (as per tests) and
-  // value won't be eligible for substitute
-  return '\x1b[' + color + 'm' + (status || EMPTY_TOKEN_VALUE_PLACEHOLDER) + '\x1b[0m'
+  return color ? '\x1b[' + color + 'm' + status + '\x1b[0m' : status
 })
 
 /**
@@ -404,9 +394,7 @@ function compile (format) {
       tokenArguments += ', ' + String(JSON.stringify(arg))
     }
 
-    return '" +\n    (' +
-      tokenFunction + '(' + tokenArguments + ') || "' + EMPTY_TOKEN_VALUE_PLACEHOLDER +
-      '") + "'
+    return '" +\n    (' + tokenFunction + '(' + tokenArguments + ') || "-") + "'
   })
 
   // eslint-disable-next-line no-new-func
--- a/index.js
+++ b/index.js
@@ -187,7 +187,7 @@ morgan.format('tiny', ':method :url :status :res[content-length] - :response-tim
  * dev (colored)
  */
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -110,12 +110,11 @@ Standard Apache common log output.
 
 ##### dev
 
-Concise output colored by response status for development use. The `:status`
-token will be colored red for server error codes, yellow for client error
-codes, cyan for redirection codes, and uncolored for all other codes.
+Concise output colored by response status for development use.
+See `:status-colored` for details.
```
dougwilson: I'm not sure this part is needed, as why jot say to see all the different tokens? We should probably either keep the color explanation here or just remove this part, I think.

index.js
```diff
@@ -260,10 +244,26 @@ morgan.token('date', function getDateToken (req, res, format) {
...
-    ? String(res.statusCode)
-    : undefined
+morgan.token('status', getStatusToken)
+
+/**
+ * colored response status code
+ */
+
+morgan.token('status-colored', function getStatusColoredToken (req, res) {
+  var status = getStatusToken(req, res)
```
dougwilson: Shouldn't the status variable here be a number and not a string?

index.js
```diff
@@ -260,10 +244,26 @@ morgan.token('date', function getDateToken (req, res, format) {
...
+  // get status color
+  var color = status >= 500 ? 31 // red
+    : status >= 400 ? 33 // yellow
+      : status >= 300 ? 36 // cyan
+        : status >= 200 ? 32 // green
+          : 0 // no color
+
+  // using placeholder here since `compile` adds it if token function returns falsy,
+  // but this coloring token returns color data anyway (as per tests) and
+  // value won't be eligible for substitute
```
dougwilson: I'm not sure why we should do this vs returning undefined and allow the compile function to take care of it. Does the actual color of the dev format change at all if we do this? If an unknown status returns something here, then this would be an outlier for how tokens currently work and likely needs explicit docs around this. But I don't actually think it's necessary, but happy to be proven wrong there :)

README.md
ryhinchey: agreed. I'll add the color explanation back in

index.js
ryhinchey: yes! I'll update it

index.js
ryhinchey: this is a great point and simplified the code :)

index.js
```diff
@@ -494,7 +508,8 @@ function headersSent (res) {
 
 function pad2 (num) {
   var str = String(num)
-
+  str = 22
```
dougwilson: I think this line here is a stray from a test you were trying out.

index.js
```diff
@@ -260,10 +244,23 @@ morgan.token('date', function getDateToken (req, res, format) {
...
+  var status = getStatusToken(req, res)
+
+  // get status color
+  var color = status >= 500 ? 31 // red
+    : status >= 400 ? 33 // yellow
+      : status >= 300 ? 36 // cyan
+        : status >= 200 ? 32 // green
+          : 0 // no color
+
+  return color ? '\x1b[' + color + 'm' + status + '\x1b[0m' : status
```
dougwilson: I'm commenting here, but this could have just as well been a comment on the docs -- I realize there is a slight quirk in the behavior of this token: technically after the token, it will change the color of everything afterwards as well. For example think of the following format string: `\x1b[47m:method :url :color-status :res[content-length]\x1b[0m`. I believe this token will end up altering the appearance of the `:res[content-length]` when compared to using `:color` token.

I will continue to think about what (if anything) we could do about that, but wanted to comment about this before I completed my thought on it in case you had any thoughts. The reason I mentioned docs is because even if we say there is nothing we can do, it should at least be documented.

index.js
ryhinchey: woops! my bad

index.js
ryhinchey: When I test this implementation I'm only seeing the status get a color.  the `'\x1b[0m'` is an ansi reset code that should reset the color.  It's used in the current `dev` format too: https://github.com/expressjs/morgan/blob/master/index.js#L202

dougwilson: Oh, sorry, your terminal may not support background colors. Try this one out: `'\x1b[4m:method :url :color-status :res[content-length]\x1b[0m'`. I can imagine the user would expect everything (but the status) to be blue.

test/morgan.js
```diff
@@ -1051,12 +1125,11 @@ describe('morgan()', function () {
       it('should color 2xx green', function (done) {
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
-          assert.strictEqual(line.substr(0, 37), '_color_0_GET / _color_32_200_color_0_')
-          assert.strictEqual(line.substr(-9), '_color_0_')
```
dougwilson: Why remove this part? This is to test that the end of the message resets the console back so non-morgan `console.log` lines are not suddenly colored.

test/morgan.js
```diff
@@ -1073,12 +1146,11 @@ describe('morgan()', function () {
       it('should color 3xx cyan', function (done) {
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
-          assert.strictEqual(line.substr(0, 37), '_color_0_GET / _color_36_300_color_0_')
-          assert.strictEqual(line.substr(-9), '_color_0_')
```
dougwilson: Why remove this part? This is to test that the end of the message resets the console back so non-morgan `console.log` lines are not suddenly colored.

test/morgan.js
```diff
@@ -1095,12 +1167,11 @@ describe('morgan()', function () {
       it('should color 4xx yelow', function (done) {
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
-          assert.strictEqual(line.substr(0, 37), '_color_0_GET / _color_33_400_color_0_')
-          assert.strictEqual(line.substr(-9), '_color_0_')
```
dougwilson: Why remove this part? This is to test that the end of the message resets the console back so non-morgan `console.log` lines are not suddenly colored.

test/morgan.js
```diff
@@ -1117,12 +1188,11 @@ describe('morgan()', function () {
       it('should color 5xx red', function (done) {
         var cb = after(2, function (err, res, line) {
           if (err) return done(err)
-          assert.strictEqual(line.substr(0, 37), '_color_0_GET / _color_31_500_color_0_')
-          assert.strictEqual(line.substr(-9), '_color_0_')
```
dougwilson: Why remove this part? This is to test that the end of the message resets the console back so non-morgan `console.log` lines are not suddenly colored.

index.js
```diff
@@ -180,30 +187,7 @@ morgan.format('tiny', ':method :url :status :res[content-length] - :response-tim
...
-
-  if (!fn) {
-    // compile
-    fn = developmentFormatLine[color] = compile('\x1b[0m:method :url \x1b[' +
-      color + 'm:status\x1b[0m :response-time ms - :res[content-length]\x1b[0m')
-  }
-
-  return fn(tokens, req, res)
-})
+morgan.format('dev', ':method :url :status-colored :response-time ms - :res[content-length]')
```
dougwilson: I think this is now entering breaking-change territory... the current 1.x `:dev` format will enture that un-ended color codes did not suddenly color the status message, but now the reset code has been removed from the start of the format.

dougwilson: PR #30 introduced the current start code for context, as it's not there by accident currently.

ryhinchey: Ok - I'll add it back! Definitely not trying to introduce a breaking change here :)

index.js
ryhinchey: yah - in this PR the `:status-colored` token resets any colors at the end so any background color disappears after `:status-colored`.   If the user is setting the background color of their log I think they'd expect that background color for the entire log and have `:status-colored` be whichever color morgan sets it to.

I think the right thing to do is remove the reset at the end of the `:status-colored` token and update `dev` format to add a reset to the end of `:status-colored`.  

That would keep the dev format as is and allow a user to do this if they wanted to add a background color:
`'\x1b[45m:method :url :status-colored \x1b[0m\x1b[45m:res[content-length]\x1b[0m'`

Do you agree?  If so, I'll update the docs

test/morgan.js
ryhinchey: sorry - added back!

test/morgan.js
ryhinchey: sorry - added back!

test/morgan.js
ryhinchey: sorry - added back!

test/morgan.js
ryhinchey: sorry - added back!

index.js
dougwilson: Yea, I'm not sure I have a preference with which way to do it, but mainly to call out in the documentation for how folks will need to use the token and what it is doing with the terminal colors.

ryhinchey: Updated the docs!

Issue #228, opened 2020-02-27
test coverage drops depending on what time the tests are run


ryhinchey:
the argument for `pad2` is generated based on the current date.  if the tests are run at a time with no single digit values for `date`, `hour`, `mins` or `secs` the test coverage will drop as the `str.length === 1` check in `pad2` is never run.

As an example, the coverage is dropping here because the tests were run when all date values were 2 digits.  If this line https://coveralls.io/builds/29010561/source?filename=index.js#L515 was ignored, the coverage would be 100%.
 
PR adding the ignore check https://github.com/expressjs/morgan/pull/229

PR #229, opened 2020-02-27
adding istanbul ignore for pad2

ryhinchey:
Related issue: https://github.com/expressjs/morgan/issues/228

I'm open to an alternative here.  I tried adding `sinon@2.0.0` (the oldest version I could install) so I could write a new test with a mocked date; however, sinon wouldn't install correctly in older versions of node https://github.com/expressjs/morgan/pull/226.

### DIFF
--- a/index.js
+++ b/index.js
@@ -495,6 +495,7 @@ function headersSent (res) {
 function pad2 (num) {
   var str = String(num)
 
+  // istanbul ignore next: num is current datetime
   return (str.length === 1 ? '0' : '') + str
 }
 


dougwilson:
I am also fine with this, and great find @ryhinchey ! I thought there were times in the past I saw the coverage drop randomly but never looked into it that hard... of course the ideal is to test it, but these are one of those things where rewriting it to expose it enough to test would (imo) break encapsulation we'd want to keep private, and some small branch like this we'll just have to skip. This won't be the first place we skip coverage in the various repos, and I think it's a good reason to.

I know that was a lot, but just wanted to provide my reassurance 👍 . I am getting a new machine set up right now and will start chewing through a few things tonight, including getting this landed.

dougwilson:
Well, the PR just failed due to a falling coverage... so maybe still need to understand better what is happening here

dougwilson:
Apparently there is a limit to the length of the ignore comment, lol

PR #232, opened 2020-04-09
updating the readme with installation and updates to format and tokens

ryhinchey:
This PR makes some updates to the readme

- add installation instructions
- link to predefined formats and tokens from first part of api docs
- add formatting to predefined format and token names in their respective lists. this makes them easier to find when browsing the docs
- move creating new tokens section to after the list of predefined tokens.  for most users who just need to use the predefined tokens, they should be able to access that content quickly without having to read about making a custom token first
- show how custom types are used in a format string

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ HTTP request logger middleware for node.js
 
 > Named after [Dexter](http://en.wikipedia.org/wiki/Dexter_Morgan), a show you should not watch until completion.
 
+## Installation
+
+```
+npm install morgan
+```
+
 ## API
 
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks for this. I would suggest best best practice is to split you pull request into logical changes an open one per change. An easy example here is adding the install section seems a separate change from some of the others.

Large pull requests are harder to review and much more likely to sit for a long time. In addition, any comment on one part of the pull request keeps the entire list of changes from landing until all parts have been resolved.

ryhinchey:
Sounds good Doug!  I’ll close this and open a few

PR #234, opened 2020-04-10
adding formatting to options, tokens, and formats for better readability

ryhinchey:


dougwilson:
Can you clarify on your thoughts on how it makes it more readable? It seems weird to add a background color to the headings, especially where the background does not extend the full page width, like a heading normally would. It also seems to deviate from the styling of all the other modules in our orgs, so if this is really desirable, we should make a plan to update them all to be consistent and not let them each drift into weirdly different styles.

ryhinchey:
> Can you clarify on your thoughts on how it makes it more readable?

When browsing the list of tokens, options, or predefined formats, I found it easier to see the individual items and then read the description of the item.  Without the formatting, it was harder to process each individual item and I found myself reading the beginning of each description before moving on to the next item.

I was reading through the multer readme which does this for their [options](https://github.com/expressjs/multer#multeropts).  I see that other repos in the express org don't format options like this.

dougwilson:
Strange, as I feel like it is harder to read them with this change, as they are no longer bold...

ryhinchey:
> Strange, as I feel like it is much easier to skip over the headings with this change, as they are no longer bold...

fair enough! i'll close this pr

Issue #222, opened 2020-02-24
Discuss: morgan@2.0.0


ryhinchey:
Opening this to start a discussion about if a v2 of morgan would be considered and what the goals of it would be.  

As someone who would be happy to contribute to the effort, I wanted to gauge interest and come up with ideas of what a v2 would be looking to accomplish.

**Some issues I think are relevant for a v2:**
- https://github.com/expressjs/morgan/issues/190 - officially remove `morgan.default`
- https://github.com/expressjs/morgan/issues/142 & https://github.com/expressjs/morgan/issues/141 - finding an alternative to adding values to the `req` and `res` objects or coming up with a safer naming convention as to not conflict with other middleware

**Some ideas:**
- ~~removing the `compile` function. the `compile` can be difficult to understand and uses what some consider to be an [unsafe technique] - (https://github.com/expressjs/morgan/blob/master/index.js#L3910)~~ - https://github.com/expressjs/morgan/issues/223 closed due to convincing by Doug that compile is valuable to users
- ~~replacing the `token` syntax with functions. Using a function is [already documented](https://github.com/expressjs/morgan#use-custom-token-formats) and if `compile` is removed, we'd need to move all tokens to plain functions.~~ - if compile stays, we need token syntax

dougwilson:
> removing the compile function. the compile can be difficult to understand and uses what some consider to be an unsafe technique

I've never heard this proposal before, but would love to hear more about it. What would someone who is making use of that feature do without it? Is there an alternative proposed here? I ask because ideally we want to improve the module, and if there is an alternative and we can put it into the 1.x line and deprecate the old functions _prior_ to suddenly dropping, that would be the most ideal path.

dougwilson:
> replacing the token syntax with functions. Using a function is already documented and if compile is removed, we'd need to move all tokens to plain functions.

I'm a little unclear on this proposal as well. All tokens are already functions. So by changing them to functions it's not clear what that would look like since they are functions today...

I think this thread is about to become very messy. Should we make a separate issue for each separate proposal so the discussions do not end up mixed together? I feel like this should just be a meta issue for _existing_ proposal we can link to and the discussion here should be on if they should be in the release vs making multiple new proposals in a single issue that would then need to be cross discussed.

Thoughts?

ryhinchey:
I think that's a great idea.  I'll open separate issues with proposals and more detail

dougwilson:
[...]

PR #238, opened 2020-04-10
move formats one level under the morgan object

ryhinchey:
This PR would close https://github.com/expressjs/morgan/issues/190.  

Per my comment https://github.com/expressjs/morgan/issues/190#issuecomment-611829994, there is not an issue with using `import` syntax in the versions of node that support it. The issue is _only_ with using import syntax via the `esm` module.

If we'd rather not land something like this under 1.x I'm  fine with that call and am all for starting the work on v2 that removes the `default` format (along with other items) as noted here https://github.com/expressjs/morgan/issues/222.

### DIFF
--- a/index.js
+++ b/index.js
@@ -145,6 +145,8 @@ function morgan (format, options) {
   }
 }
 
+morgan.formats = {}
+
 /**
  * Apache combined log format.
  */
@@ -162,7 +164,7 @@ morgan.format('common', ':remote-addr - :remote-user [:date[clf]] ":method :url 
  */
 
 morgan.format('default', ':remote-addr - :remote-user [:date] ":method :url HTTP/:http-version" :status :res[content-length] ":referrer" ":user-agent"')
-deprecate.property(morgan, 'default', 'default format: use combined format')
+deprecate.property(morgan.formats, 'default', 'default format: use combined format')
 
 /**
  * Short format.
@@ -452,7 +454,7 @@ function createBufferStream (stream, interval) {
  */
 
 function format (name, fmt) {
-  morgan[name] = fmt
+  morgan.formats[name] = fmt
   return this
 }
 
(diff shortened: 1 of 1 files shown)


dougwilson:
I can certainly say that it is not possibe actually land this on 1.x because it moves publically accessible API; the format function are now no longer available within the other tokens as the same name, for example.

I'm just noting that if that affects your desire for this PR, as it would be a v2 PR as-is, and it sounded like you had a different change in mind for v2.

ryhinchey:
> I can certainly say that it is not possibe actually land this on 1.x because it moves publically accessible API; the format function are now no longer available within the other tokens as the same name, for example.

 I viewed this as an internal change that didn't effect the public facing api (what's available via `module.exports`).  If someone is using morgan per the documented api, this shouldn't be a breaking change, right?

Going forward, it's good to understand what constitutes a breaking change.  You're saying that because someone _can_ access these directly on the morgan object and with this PR they'd no longer be able to, that it's a breaking change?

If that's the case (and I'm fine with it), I'll close this PR and make a PR adding tests that would prevent this PR from being made in the first place.

dougwilson:
> I viewed this as an internal change that didn't effect the public facing api (what's available via module.exports). If someone is using morgan per the documented api, this shouldn't be a breaking change, right?

What is on the `module.exports` is what is passed into your custom `format` function. This is documented public API, but docs can always be improved if needed. Here is an example app that would be broken with the PR here, which isn't doing anything non-public IMO:
```js

var app = express()

  return someCondition() ? tokens.default(tokens, req, res) : tokens.dev(tokens, req, res)
}))

app.get('/', function (req, res) {
  res.send('hello, world!')
})
```

ryhinchey:
> This PR is directly altering the names of exports of the module, which seems to be a pretty straight-forward violation.

I took a look at `module.exports` and since I didn't change anything there, I figured this wouldn't be a breaking change.  You're right though - given I'm changing the properties on the `morgan` object, this is indeed a breaking change.

Thanks for giving a thoughtful review here. It's very helpful!

I'm going to close this PR and submit a new one that adds tests that would have not passed for this PR.

PR #241, opened 2020-04-18
add chunk and chunk size indicator

### LINKED ISSUE #102: Size of response with no 'content-length' header
Is it possible for a tag to be added that uses 'content-length' is the type of the data sent is not in chunks and a chunk size indicator when the data is sent in chunks?

mum-never-proud:
resolves #102

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,6 +109,14 @@ function morgan (format, options) {
     // response data
     res._startAt = undefined
     res._startTime = undefined
+    res._chunkSize = 0
+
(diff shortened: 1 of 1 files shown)


mum-never-proud:
@dougwilson i am not so sure that this is a great idea is correct so i haven't started writing test-cases, 

any pointer on how we can calculate the size of response body?
i don't think that the transfer type would be chunked if there's no Content-Length

thanks for your patience 💪

PR #252, opened 2020-12-21
Add an option to print logs to console..

Zorono:


dougwilson:
Hi @Zorono ! Writing the logs to the console is the default behavior.

Zorono:
> Hi @Zorono ! Writing the logs to the console is the default behavior.

LOL... it seems that I am blind since I can't find anything that makes it writes to the console except **debug** pkg...

PR #255, opened 2021-08-16
Added a predefined format for JSON logging.

ziadkhouri:
Have added a predefined format for JSON logging. Follows the [Google Cloud Logging](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) specification

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The minimal output.
 
 ##### json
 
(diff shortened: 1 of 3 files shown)


dougwilson:
This seems specific to a sertain cloud. You can always define that format for your app as needed, no need to add it here directly.

PR #263, opened 2022-06-29
Fix semantics in README

permalik:
Updated README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-# morgan
 
 [![NPM Version][npm-version-image]][npm-url]
 [![NPM Downloads][npm-downloads-image]][npm-url]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,3 @@
-# morgan
```
dougwilson, line 1: Why are you remocing the module name from the top? This is how we format all our expressjs readmes.

README.md
```diff
@@ -7,7 +6,11 @@
 
 HTTP request logger middleware for node.js
 
-> Named after [Dexter](http://en.wikipedia.org/wiki/Dexter_Morgan), a show you should not watch until completion.
+// I thought you meant: "The latter seasons of Dexter are no good and should be skipped"
```
dougwilson, line 9: I'm not sure what all these added to the readme do. It seems to be javascript comments, but markdown just will render as plain text in the readme.

dougwilson (changes requested):

PR #284, opened 2023-10-11
doc: add replace reference from CJS to EJS 

green-veggies:
To import this we have to use ```var morgan = require("morgan");``` which is outdated
we should use ``` import morgan from "morgan"; ```


bjohansebas:
This package still supports very old versions of Node.js. We have updated the syntax in the packages that have modern Node.js support.

PR #288, opened 2024-04-10
Update readme.md

adrinorosario:
Added new ES6 module import for importing module


UlisesGascon:
This is duplicated, see #284

PR #289, opened 2024-04-23
Update README.md

harishghasolia07:
Update the way to import method. Updated by common JS to EJS


UlisesGascon:
Thanks for your contribution, but this PR seems duplicated (see: #284)

Issue #302, opened 2025-05-29
`morgan('dev')` does not respect `NO_COLOR=1` environment variable


MaddyGuthridge:
The `NO_COLOR` environment variable is the widely-accepted standard for disabling colour in terminal applications. While most users (myself included) ***love*** colourful outputs, these outputs can cause significant accessibility issues for people with limited vision.

A wider explanation can be found at [no-color.org](https://no-color.org/)

As such, Morgan should detect the use of the `NO_COLOR` environment variable, and disable coloured output when it is set.

## Environment information

**Version**:
<!-- 
To find the installed version of a package, you can check the `package.json` file in the root directory of your project. The version will be listed under `dependencies` or `devDependencies`, like this:

```json
"dependencies": {
  "express": "4.18.2"
}
-->

* `"morgan": "^1.10.0"`
* `"express": "^4.19.2"`

**Platform**:
<!-- 
UNIX: output of uname -a
Windows: output of "$([Environment]::OSVersion.VersionString) $(('x86', 'x64')[[Environment]::Is64BitOperatingSystem])" in PowerShell console 
-->

`Linux yzma 6.14.6-300.fc42.x86_64 #1 SMP PREEMPT_DYNAMIC Fri May  9 20:11:19 UTC 2025 x86_64 GNU/Linux`

**Node.js version**:
<!-- 
Output of node -v.
-->

v20.18.1

**Any other relevant information**:

## What steps will reproduce the bug?

Run a server passing the environment variable `NO_COLOR=1`

jonchurch:
Shipped in [1.12.0](https://github.com/expressjs/morgan/releases/tag/1.12.0)

Issue #265, opened 2022-07-27
Adding a token named token causes all subsequent tokens to break.


WadeDutch:
This works just fine:
var logger = require('morgan');
logger.token('ip', function (req, res) { return req.ip ? req.ip : "" });
logger.token('uuid', function (req, res) { return (req.token ? req.token.uuid : "") });
logger.token('token', function (req, res) { return (req.token ? req.token.token : "") });

This gives "TypeError: tokens.uuid is not a function" when called in the logger:
var logger = require('morgan');
logger.token('ip', function (req, res) { return req.ip ? req.ip : "" });
logger.token('token', function (req, res) { return (req.token ? req.token.token : "") });
logger.token('uuid', function (req, res) { return (req.token ? req.token.uuid : "") });

dougwilson:
Ah yes, that is def a footgun. We should make it an error when trying to use a reserved word in 1.x line and also get a beta of 2.0 published which allows any token name and the other 2.0 stuff, haha.

WadeDutch:
Figured as much. Is there a list of reserved words anywhere?

dougwilson:
Not currently, so that is another thing to add: docs for this :)

Samiullah-2004:
Hi, I'd like to work on this issue. I'll add a check that throws a clear error when a reserved token name (like "token") is used, instead of silently breaking.

PR #303, opened 2025-05-29
A11y: support NO_COLOR environment variable in output

MaddyGuthridge:
Resolves #302

This is a quick fix, rather than a major change to the system Morgan uses for styled console output. In an ideal world, using `NO_COLOR` would disable all styling of output, which would remove the presence of control characters in things like CI logs. Currently, this PR just sets the color to `0` if `NO_COLOR` is set in `process.env`, which is acceptable for accessibility.

As a side note: I'm not used to writing code for such old versions of Node, so apologies if I made any mistakes that break that compatibility. I tried testing using Node 0.8.28, but was unable to get `npm i` to work correctly. Let me know if I broke anything and I'll do my best to fix it up :)

- [x] Implementation
- [x] Test suite passes with 100% coverage
- [x] Linting passes with no errors or warnings

I accept the certificate of origin, and am happy for my contribution to use the project's MIT software license.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,6 +41,14 @@ var CLF_MONTH = [
 ]
 
 /**
+ * Whether to disable colored output in dev mode.
+ * Accessibility standard as per https://no-color.org/
+ * @private
+ */
+
+var NO_COLOR = !!process.env.NO_COLOR;
+
+/**
  * Default log buffer duration.
  * @private
  */
@@ -187,11 +195,12 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
     : undefined
 
   // get status color
-  var color = status >= 500 ? 31 // red
-    : status >= 400 ? 33 // yellow
-      : status >= 300 ? 36 // cyan
-        : status >= 200 ? 32 // green
-          : 0 // no color
+  var color = NO_COLOR ? 0 // color disabled by environment variable
+    : status >= 500 ? 31 // red
+      : status >= 400 ? 33 // yellow
+        : status >= 300 ? 36 // cyan
+          : status >= 200 ? 32 // green
+            : 0 // no color
 
   // get colored function
   var fn = developmentFormatLine[color]
--- a/index.js
+++ b/index.js
@@ -41,14 +41,6 @@ var CLF_MONTH = [
 ]
 
 /**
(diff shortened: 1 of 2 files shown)


krzysdz (commented):
I believe that `NO_COLOR=1` should **entirely remove the colour-related ANSI escape codes** instead of emitting the _reset_ sequence.

MaddyGuthridge:
Makes sense to me, although that would require a bit of a bigger refactor. I am pretty flat-out right now so probably won't have time to sort that out for a few weeks at least. I'll keep it on my todo list though.

krzysdz:
I have an idea that may be a little weird, but should work - after the current `'dev'` format (in [line 233](https://github.com/expressjs/morgan/blob/bd171d7ed3df7b1b6311f3c202e237aebe1bf830/index.js#L233)) add something like

```js
/**
 * NO_COLOR override of dev
 */

if (process.env.NO_COLOR) {
  morgan.format('dev', ':method :url :status :response-time ms - :res[content-length]')
}
```

It assumes that the `NO_COLOR` env var will be set before requiring morgan, which should be the case most of the time (this may cause problems with tests, as the relevant tests would have to import morgan separately). This should work, because `morgan.format()` performs a simple assignment, so calling it with the same name overwrites the previous format (this causes some problems as `.format()` and `.token()` can overwrite **all** `morgan` properties - #265):

https://github.com/expressjs/morgan/blob/bd171d7ed3df7b1b6311f3c202e237aebe1bf830/index.js#L480-L491

Alternatively, the `developmentFormatLine` and the string with no escape sequences could be stored as variables and set as the `dev` format when instantiating morgan somewhere above

https://github.com/expressjs/morgan/blob/bd171d7ed3df7b1b6311f3c202e237aebe1bf830/index.js#L107-L110

This would make testing easier, but IMO would negatively impact the code structure (suddenly one format is defined inside `morgan()`, not with other formats, and uses separate variables).

@jonchurch @UlisesGascon do you think this would be an acceptable solution?

UlisesGascon:
Based on https://github.com/expressjs/morgan/pull/303#issuecomment-5349560767, I think that https://github.com/expressjs/morgan/pull/372 supperseeds this pr :thinking:

PR #339, opened 2026-03-10
fix: respect NO_COLOR environment variable

abhii734:
## Bug\n\nmorgan dev format does not respect NO_COLOR=1 environment variable (morgan#302)\n\n## Fix\nCheck process.env.NO_COLOR and disable colored output when set.\n\nThis complies with no-color.org standard for accessibility.


krzysdz:
Duplicate of #303

PR #340, opened 2026-03-10
fix: resolve #265 - custom token named token breaks subsequent tokens

### LINKED ISSUE #265: Adding a token named token causes all subsequent tokens to break.

abhii734:
## Bug (morgan#265)

Adding a token named "token" causes all subsequent tokens to break. This is because morgan.token() overwrites the morgan.token function itself.

## Fix
- Store custom tokens in separate tokenStore object
- Prevent overwriting built-in properties (token, format, compile)
- Add morgan.getToken() method for compiled format functions to lookup tokens
- Also add NO_COLOR support in dev format (morgan#302)

## Verification
Test code that previously failed:
```js
var logger = require("morgan");
logger.token("ip", function (req, res) { return req.ip ? req.ip : "" });
logger.token("token", function (req, res) { return (req.token ? req.token.token : "") });
logger.token("uuid", function (req, res) { return (req.token ? req.token.uuid : "") });
```
This now works without "TypeError: tokens.uuid is not a function"

### DIFF
--- a/index.js
+++ b/index.js
@@ -180,18 +180,19 @@ morgan.format('tiny', ':method :url :status :res[content-length] - :response-tim
  * dev (colored)
  */
 
-morgan.format('dev', function developmentFormatLine (tokens, req, res) {
+morgan.format('dev', function developmentFormatLine (morgan, req, res) {
   // get the status code if response written
   var status = headersSent(res)
     ? res.statusCode
     : undefined
 
-  // get status color
-  var color = status >= 500 ? 31 // red
-    : status >= 400 ? 33 // yellow
-      : status >= 300 ? 36 // cyan
-        : status >= 200 ? 32 // green
-          : 0 // no color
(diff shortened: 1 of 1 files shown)


UlisesGascon:
Closing in favor of https://github.com/expressjs/morgan/pull/359

PR #341, opened 2026-03-29

mahmoodhamdi:
## Summary

The `dev` format now respects the `NO_COLOR` environment variable by omitting ANSI escape codes when it is set.

## Problem

The `dev` format uses ANSI color codes to colorize HTTP status codes (green for 2xx, cyan for 3xx, yellow for 4xx, red for 5xx). However, it does not respect the [`NO_COLOR`](https://no-color.org/) environment variable, which is the widely-accepted standard for disabling color output in terminal applications.

This causes accessibility issues for users with limited vision and problems in environments where ANSI escape codes are not supported or cause garbled output (CI logs, file redirection, etc.).

## Fix

When `NO_COLOR` is set in `process.env`, the `dev` format outputs plain text without any ANSI escape sequences:

```
GET / 200 2.341 ms - 13
```

Instead of:

```
\x1b[0mGET / \x1b[32m200\x1b[0m 2.341 ms - 13\x1b[0m
```

The no-color format is cached separately from colored formats, so there is no performance impact.

## Testing

Added a test that verifies the output contains no ANSI escape codes when `NO_COLOR` is set. All 83 tests pass. Lint clean.

Fixes #302

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,6 +186,18 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
     ? res.statusCode
     : undefined
 
+  // respect NO_COLOR environment variable (see https://no-color.org)
+  if ('NO_COLOR' in process.env) {
+    var fn = developmentFormatLine.nocolor
+
+    if (!fn) {
+      fn = developmentFormatLine.nocolor = compile(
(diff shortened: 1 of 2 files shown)


krzysdz:

PR #342, opened 2026-03-30
refactor: replace deprecated .substr() with .substring()

singhankit001:
String.prototype.substr() is deprecated per MDN. This PR replaces all occurrences in the test file with the standard .substring() method.


jonchurch:
Search replacing `substr` to `substring` does not work, they have different APIs. So the tests fail w/ this PR.

The deprecation is about browsers, not node. There's different APIs to do this today, but `substr` works fine here and doesn't need to be fixed.

PR #344, opened 2026-04-15
support NO_COLOR environment variable in dev format

SAY-5:
The dev format now checks for the NO_COLOR environment variable
(https://no-color.org/) and omits ANSI escape codes when it is set.

When `NO_COLOR` is present in `process.env`, the dev format compiles a
plain-text format string without any `\x1b[` sequences and caches it
under a separate `'nocolor'` key so colored and uncolored output don't
interfere with each other.

Two new tests cover the behavior:
- output contains no ANSI escape codes when `NO_COLOR` is set
- output still includes the HTTP method and status code

Fixes #302

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,20 +186,27 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
     ? res.statusCode
     : undefined
 
-  // get status color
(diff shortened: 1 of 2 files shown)


krzysdz:

PR #345, opened 2026-04-15
fix: preserve request start time across multiple morgan instances

### LINKED ISSUE #141: Multiple morgan instances stubs out original start time
Hello, I was looking to have multiple logs, the standard `process.stdout`, a `fileStreamRotate` and a `mongo` variant. I was looking to see if there would be any issues with a few being synchronous and a few being asynchronous. I didn't see much in the way of problems except the fact that the _startTime is set to undefined each time. Which could cause problems between multiple morgan instances.

I understand what I am about to write is an exaggeration and prob not recommended anyway, but I just wanted to point it out.

```js
var morgan = require("morgan");
var app = require("express")();

var stream1 = require("fs").createWriteStream("log.log");
var stream2 = process.stdout;

app.use(morgan("dev", { stream: stream1 }));
app.use((req, res, next) => { setTimeout(next, 2000); }); // Wait 2 seconds
app.use(morgan("dev", { stream: stream2 }));
app.use((req, res) => { res.send(); });

app.listen(3000);
```

As I said, this would be an extremely strange behaviour, but it does point out that both morgans record that the response time is ~8ms rather than over 2008ms

- Colin

SAY-5:
When multiple morgan middleware instances are chained together, the second
one resets req._startAt and res._startAt before recording new values. This
means all instances end up measuring response time from when the last morgan
middleware ran, not from the actual request arrival.

The fix only records start timestamps once per request/response pair. If
_startAt is already set (by an earlier morgan instance), it is left alone.

Before (two morgan instances with a 100ms delay between them):
```
0.998
0.998
```

After:
```
106.552
106.552
```

Both loggers now correctly report the full response time from request start.

Includes a test that chains two morgan instances with a 50ms delay and
verifies both report times >= the delay.

Fixes #141

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,16 +102,12 @@ function morgan (format, options) {
 
   return function logger (req, res, next) {
     // request data
-    req._startAt = undefined
-    req._startTime = undefined
     req._remoteAddress = getip(req)
 
-    // response data
-    res._startAt = undefined
-    res._startTime = undefined
-
-    // record request start
-    recordStartTime.call(req)
+    // record request start (only once, even with multiple morgan instances)
+    if (!req._startAt) {
+      recordStartTime.call(req)
+    }
 
     function logRequest () {
       if (skip !== false && skip(req, res)) {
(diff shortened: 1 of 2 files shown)


SAY-5:
Anything I can do to move this along? The fix keeps the start time on the request object so a second morgan instance in the same chain doesn't reset it, and CI is green.

PR #348, opened 2026-05-31
feat: respect NO_COLOR environment variable in dev format

xxiaoxiong:
## Problem

The `morgan('dev')` format always outputs ANSI color codes, with no way to disable them. The `NO_COLOR` environment variable (https://no-color.org/) is the widely-accepted standard for disabling colour in terminal applications, but morgan does not check for it.

## Changes

Added a `NO_COLOR` environment variable check at the start of the `developmentFormatLine` function. When `process.env.NO_COLOR` is set, the dev format outputs plain text without ANSI color codes, matching the format of the `tiny` or `short` format but preserving the token structure of `dev`.

## Why

Colorful outputs can cause significant accessibility issues for people with limited vision. The `NO_COLOR` standard is supported by many CLI tools and frameworks, and this change brings morgan in line with the ecosystem.

Fixes #302

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,6 +186,17 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
     ? res.statusCode
     : undefined
 
+  // respect NO_COLOR environment variable (https://no-color.org/)
+  if (process.env.NO_COLOR !== undefined) {
+    var fn = developmentFormatLine._nocolor
+
+    if (!fn) {
+      fn = developmentFormatLine._nocolor = compile(':method :url :status :response-time ms - :res[content-length]')
+    }
+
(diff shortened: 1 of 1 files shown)


Exagone313:
Hello,

This user is a bot that is automatically generating pull requests on a lot of projects.

This could be used to gain access to insecure GitHub Actions setups. By default, approval on pull requests is required only for first-time contributors. If that is the case for this project, you might need to change the setting `Approval for running fork pull request workflows from contributors`.

I recommend closing this PR.

<https://github.com/search?q=author%3Axxiaoxiong+is%3Apr&s=created&type=pullrequests>

Thanks.

PR #350, opened 2026-06-02
Release 1.11.0

jonchurch:
1.11.0 / 2026-06-02
===================
  * add `:pid` token

  Security Fix:
  * Escape control characters in `:remote-user` token to prevent log injection 
    * Fixes [CVE-2026-5078](https://www.cve.org/CVERecord?id=CVE-2026-5078) [GHSA-4vj7-5mj6-jm8m](https://github.com/expressjs/morgan/security/advisories/GHSA-4vj7-5mj6-jm8m)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,10 @@
-unreleased
+1.11.0 / 2026-06-02
 ===================
   * add `:pid` token
+
+  Security Fix:
+  * Escape control characters in `:remote-user` token to prevent log injection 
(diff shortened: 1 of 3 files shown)


UlisesGascon (approved):

jonchurch:
for some reason npm access is not cache busting? as in, I just added myself to the package team in npm org, but my access hasn't propogated (we keep teams at the express org level, so we add ourselves to a package only when we need to publish)

So @UlisesGascon is stepping in to run publish, we are in our release call rn

PR #353, opened 2026-06-24

luisangelrod:
## Summary

The `dev` format currently emits ANSI color escape codes unconditionally. This change follows the [`NO_COLOR`](https://no-color.org/) convention: color is disabled when `NO_COLOR` is present and non-empty, while an empty value preserves the existing behavior.

## Changes

**`index.js`**

- Check whether `NO_COLOR` has a non-empty value before compiling the dev format.
- Cache a plain format without ANSI escape codes when color is disabled.
- Preserve the current colored output when `NO_COLOR` is absent or empty.

**`test/morgan.js`**

- Cover color-free 2xx and 5xx output with `NO_COLOR=1`.
- Cover the convention's empty-value behavior.
- Isolate the dev-format tests from an inherited `NO_COLOR` shell value and restore the original environment after each test.

## Testing

- `npm test` — 91 passing
- `npm run lint`
- `git diff --check`

Closes #302

### DIFF
--- a/index.js
+++ b/index.js
@@ -219,13 +219,19 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
         : status >= 200 ? 32 // green
           : 0 // no color
 
+  // respect NO_COLOR environment variable (https://no-color.org/)
+  var noColor = 'NO_COLOR' in process.env
+
   // get colored function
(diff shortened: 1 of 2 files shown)


krzysdz:

PR #359, opened 2026-07-06
fix: protect method names from token registration collisions

baima365-web:
## Problem

Calling `morgan.token('token', fn)` overwrites the `morgan.token()` method itself, breaking all subsequent token registrations. This is because `morgan` serves as both the module export (with methods `token`, `format`, `compile`) and the token registry.

Example:
```js
var logger = require('morgan');
logger.token('token', function (req, res) { return 'value' });
// morgan.token is now the custom function, not the registration method
logger.token('uuid', function (req, res) { return 'uuid' }); // TypeError: tokens.uuid is not a function
```

## Fix

Make the `token`, `format`, and `compile` methods non-writable using `Object.defineProperties`. This prevents user token registrations from overwriting the module API while maintaining full backward compatibility.

## Test plan

- [x] Existing built-in tokens work correctly
- [x] Custom token registration works
- [x] Compiled format functions resolve tokens correctly
- [x] Middleware logs correctly
- [x] Registering reserved names (token, format, compile) now throws instead of silently breaking

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,6 +19,13 @@ module.exports.compile = compile
 module.exports.format = format
 module.exports.token = token
 
+// Protect method names from being overwritten by token registrations
+Object.defineProperties(module.exports, {
+  compile: { writable: false, configurable: false },
+  format: { writable: false, configurable: false },
+  token: { writable: false, configurable: false }
+})
+
 /**
  * Module dependencies.
  * @private


UlisesGascon (approved):
LGTM! WDYT @jonchurch @bjohansebas ?

bjohansebas (approved):
LGTM, should we add tests?

krzysdz (commented):
While this prevents accidentally overwriting important functions, there are still some rather unexpected behaviours:

1. There is completely no information that the given token or format cannot be defined (this will become clear if someone tries to use it, but may be hard to debug).
2. Formats and tokens can overwrite each other. While overwriting a token with a different token with the same name or overwriting a format with a different format using the same name (like I have previously suggested in https://github.com/expressjs/morgan/pull/303#issuecomment-5349560767, so #377 now depends on this behaviour) is something that IMO should be possible, format-token collisions should not happen.

Would it be a breaking change if formats and tokens were not defined directly as `morgan` properties, but instead lived separately as `morgan.tokens` and `morgan.formats` or something similar? This would allow using any names (no conflicts with the morgan functions) and get rid of collisions.

https://github.com/expressjs/morgan/blob/51007f94720168a608dfd7c4f61d706bc636a0ef/index.js#L492-L495

https://github.com/expressjs/morgan/blob/51007f94720168a608dfd7c4f61d706bc636a0ef/index.js#L579-L582

UlisesGascon:
I think that https://github.com/expressjs/morgan/pull/385 is a more solid approach but does not solve all the concerns from https://github.com/expressjs/morgan/pull/359#pullrequestreview-5045789297. Also I am considering this as semver-minor based on https://github.com/expressjs/morgan/issues/265#issuecomment-1197363886 but maybe that requires a review or split between 1.x and 2.x goals

PR #360, opened 2026-07-07

armorbreak001:
## Problem

The `dev` format always outputs ANSI-colored text, ignoring the widely-adopted `NO_COLOR` environment variable (https://no-color.org/).

This causes accessibility issues for users with limited vision who may have difficulty reading colored terminal output.

## Fix

When `NO_COLOR` is set (even to an empty string), the `dev` format outputs plain text without ANSI escape codes.

## Changes

- Modified the `dev` format function in `index.js` to check `process.env.NO_COLOR`
- When set, compiles a plain-text format instead of a colored one
- Cache key includes a `-nc` suffix to separate plain-text and colored compiled formats

## Testing

- All 88 existing tests pass
- Manual test:
  ```bash
  NO_COLOR=1 node -e "require('./index.js')('dev')" | cat -v | grep '\\x1b'
  # Should return nothing (no ANSI codes)
  ```

Fixes #302

### DIFF
--- a/index.js
+++ b/index.js
@@ -212,6 +212,9 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
     ? res.statusCode
     : undefined
 
+  // check for NO_COLOR environment variable (https://no-color.org)
+  var noColor = process.env.NO_COLOR != null
+
   // get status color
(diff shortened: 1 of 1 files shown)


krzysdz:

PR #371, opened 2026-08-19
Fix: respect NO_COLOR env var in dev format (#302)

vjymisal0:
Resolves #302 by detecting the standard NO_COLOR environment variable. If set, the dev format will use an uncolored version of the compile string.


krzysdz:

PR #377, opened 2026-08-27
feat: NO_COLOR env var removes color in dev format

jonchurch:
My approach to solve #302 

Taking a dep on the env for the behavior of dev format is the whole feature, which required a little hardening of the test setup to ensure that _people for whom this feature is intended, users of `NO_COLOR`_ can still run the normal test suite locally without it failing due to `NO_COLOR` failing the main suite's dev tests.

Because mocha doesnt support env or process isolation well (a second file in the same mocha invocation shares `process.env` across all files) and my approach favored checking NO_COLOR once at module require, I added a second mocha invocation to the main `test` npm script.

This shouldnt affect nyc coverage, it should still merge it all `¯\_(ツ)_/¯`

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,11 @@ for information codes.
 GET /dev 200 0.224 ms - 2
 ```
 
+Coloring is disabled when the [`NO_COLOR`](https://no-color.org) environment
+variable is set to any non-empty value, in which case the output above is
+written without any escape sequences. The variable is read once, when `morgan`
+is first required.
+
 ##### short
 
 Shorter than default, also including response time.
--- a/index.js
+++ b/index.js
@@ -213,17 +213,6 @@ morgan.format('tiny', ':method :url :status :res[content-length] - :response-tim
  */
 
 morgan.format('dev', function developmentFormatLine (tokens, req, res) {
-  // when NO_COLOR is set, use an escape free variant of dev format
-  if (NO_COLOR) {
-    var noColorFn = developmentFormatLine.noColor
-
-    if (!noColorFn) {
-      noColorFn = developmentFormatLine.noColor = compile(':method :url :status :response-time ms - :res[content-length]')
-    }
-
-    return noColorFn(tokens, req, res)
-  }
-
   // get the status code if response written
   var status = headersSent(res)
     ? res.statusCode
@@ -247,6 +236,12 @@ morgan.format('dev', function developmentFormatLine (tokens, req, res) {
 
   return fn(tokens, req, res)
 })
+
+// NO_COLOR (https://no-color.org): when set and not an empty string, the dev
+// format is replaced at load time with a variant free of escape sequences
+if (NO_COLOR) {
+  morgan.format('dev', compile(':method :url :status :response-time ms - :res[content-length]'))
+}
 
 /**
  * request url
--- a/index.js
+++ b/index.js
@@ -47,6 +47,8 @@ var CLF_MONTH = [
 
 var DEFAULT_BUFFER_DURATION = 1000
 
+var NO_COLOR = Boolean(process.env.NO_COLOR)
+
 /**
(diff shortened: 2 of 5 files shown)


jonchurch:
I went way down a rabbit hole on this one tbh.

The original PR was mergeable, but still I didnt like the read of process.env on every log, it almost triples the perf cost (we're talking like ~200 nanoseconds to ~600 nanoseconds on an already only for dev format, so not really relevant, but still, and the cost was worse for users w/ it unset bc thats how process.env reads work). And the fact that the PR would fail tests for anyone who presumably uses NO_COLOR themeslves just seemed a little tragic. 

We got like 10 autoamted PRs for this one issue or something, and I dont know that any even addressed these small issues, but I feel good about the time spent and taking this one over.

krzysdz (commented):
If we're going to [check `NO_COLOR` on module import](https://github.com/jonchurch/morgan/blob/3490efd9870a54b7d55e47eada4afbecd2961f64/index.js#L50), maybe we could just "overwrite" the dev format instead of adding a condition that is executed every time (something that I suggested in https://github.com/expressjs/morgan/pull/303#issuecomment-5349560767).

jonchurch:
That's a good approach @krzysdz, I got too distracted by meeting what was already in there, ill push up a change

krzysdz:
If morgan didn't have to support old Node.js versions we could use some functions from `node:util`/`node:tty`:

- [`util.styleText()`](https://nodejs.org/docs/latest/api/util.html#utilstyletextformat-text-options): detects `NO_COLOR`, `FORCE_COLOR`, `NODE_DISABLE_COLORS` and support for displaying colours. Unfortunately, it would probably have worse performance as it checks those on every function call and requires Node.js v20.18.0/v22.8.0 (or v26.1.0 for `#RRGGBB`)
- [`util.stripVTControlCharacters()`](https://nodejs.org/docs/latest/api/util.html#utilstripvtcontrolcharactersstr): just strips all control characters (not really necessary, especially since the user input is mostly sanitised - https://github.com/expressjs/morgan/commit/b3f5d9bdb388690dfae9c06ab966328f49b7982b), requires Node.js v16.11.0
- [`writeStream.hasColors()`](https://nodejs.org/api/tty.html#writestreamhascolorscount-env) (as `process.stdout.hasColors()`): respects the env variables and checks if the terminal supports the requested number of colours (like `util.styleText()`). Probably **the most useful**, but requires Node.js v10.16.0

krzysdz (approved):

jonchurch:
weirdly the ecosystem is not even very consistent w/ NO_COLORS, Node's builtins dont implement the `NO_COLORS=''`  empty string case where it's supposed to only turn off colors "when present and not an empty string". That part was maybe added to the informal "spec" later after some implemented their behavior? (yeah, 2022 https://github.com/jcs/no_color/commit/99f90e27d0fac6a34d07998f8f577e1f2f620c61)

If we land it like this, compliant with the text, then we're actually noncompliant with node's built in tools. Im down to merge this and not think about it again though.

krzysdz:
> weirdly the ecosystem is not even very consistent w/ NO_COLORS, Node's builtins dont implement the `NO_COLORS=''` empty string case where it's supposed to only turn off colors "when present and not an empty string". That part was maybe added to the informal "spec" later after some implemented their behavior?
> 
> If we land it like this, compliant with the text, then we're actually noncompliant with node's built in tools. Im down to merge this and not think about it again though.

`writeStream.hasColors()` on v26.8.1 checks if the `NO_COLOR` string is non-empty

Node.js shell showing process.stdout.hasColors called with no env (true), NO_COLOR as empty string (true) and NO_COLOR set to string 0 (false)

It would be really funny if Node.js used different (and not compatible) ways of checking it in multiple places.

UlisesGascon (approved):

PR #387, opened 2026-09-20
fix: keep per-logger start times for multiple morgan instances

### LINKED ISSUE #141: Multiple morgan instances stubs out original start time

dyk1454683243-sudo:
Fixes #141.

## Problem

Each `morgan()` middleware writes request/response start times onto the shared `req`/`res` objects (`_startAt` / `_startTime`) and clears those fields when it runs. With two loggers and work between them, the later instance stomps the earlier one, so **both** `:response-time` / `:total-time` tokens measure from the later logger.

Reproduction (from the issue, delay shortened):

```js
app.use(morgan('dev', { stream: stream1 }))
app.use(function (req, res, next) { setTimeout(next, 80) })
app.use(morgan('dev', { stream: stream2 }))
```

Before this change both loggers reported ~0.2ms. The first logger should include the delay (~80ms).

## Approach

Follow maintainer guidance on #141: each morgan instance records what *it* saw, and a later logger must not overwrite an earlier one.

- Snapshot start times in the middleware closure (per logger, per request)
- Restore that snapshot when formatting the log line so built-in and custom tokens still read `req._startAt` / `res._startAt`
- If the caller `delete`s `_startAt` (existing hidden-property tests), leave it deleted so those tokens stay empty
- Keep ES5 / Node 0.8 syntax

This is **not** the "never reset, share one start time" variant. That would make the later logger also report the full delay (a behavior change dougwilson called out as major-worthy). Here:

- first logger: time from when it ran (includes the inter-logger delay)
- second logger: time from when *it* ran (does not include that delay)

Two loggers mounted back-to-back with no delay still report essentially the same short times.

## Test plan

- [x] Reproduced #141 with two morgan instances and a delay (both ~0.2ms before; first ~80ms / second ~0.2ms after)
- [x] Regression tests for `:response-time` and `:total-time` with two loggers + delay
- [x] Sanity test that two loggers mounted together both emit a response-time
- [x] Existing hidden-property / immediate / after-response-sent tests still pass
- [x] `npm test` — 98 + 6 passing
- [x] `npm run lint` — clean


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

